PR #9, opened 2019-05-21
Teach orka to stop server

nikostoulas:
shuttdown

### DIFF
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -13,6 +13,8 @@ import riviere from './initializers/koa/riviere';
 import addRequestId from './initializers/koa/add-request-id';
 import _defaults from './default-options';
 import { OrkaOptions } from './typings/orka';
+import assert = require('assert');
+import { Server } from 'http';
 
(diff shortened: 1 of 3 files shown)


src/builder.ts
```diff
@@ -36,6 +38,7 @@ const builder = (defaults: Partial<OrkaOptions> = _defaults) => {
   const middlewares: Middleware<any>[] = [];
 
   const _ = {
+    server: null as Server,
```
anastasios-workable: Do you want to expose that to the clients? I would suggest moving this to L38 instead.

nikostoulas: Done.

src/default-options.ts
```diff
@@ -15,5 +15,6 @@ export default {
   },
   beforeMiddleware: [],
   afterMiddleware: [],
-  beforeStart: []
+  beforeStart: [],
+  initialize: null
```
nikostoulas: @anastasios-workable This overlaps with beforeStart. Initialize runs after diamorphosis while beforeStart is another queued function that is run after middlewares are initialized. Wdyt?

PR #10, opened 2019-05-21
Add initialize method

nikostoulas:
initialize-script

### DIFF
--- a/src/default-options.ts
+++ b/src/default-options.ts
@@ -15,5 +15,6 @@ export default {
   },
   beforeMiddleware: [],
   afterMiddleware: [],
-  beforeStart: []
+  beforeStart: [],
+  initialize: null
 };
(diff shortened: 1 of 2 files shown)


src/default-options.ts
```diff
@@ -15,5 +15,6 @@ export default {
   },
   beforeMiddleware: [],
   afterMiddleware: [],
-  beforeStart: []
+  beforeStart: [],
+  initialize: null
```
nikostoulas: @anastasios-workable This overlaps with beforeStart. Initialize runs after diamorphosis while beforeStart is another queued function that is run after middlewares are initialized. Wdyt?

nikostoulas: TLDR I ported orka into cerberus and beforeStart was not enough

anastasios-workable: I really think that diamorphosis should be moved out, it complicates things a lot and is not that much boilerplate tbh. Thoughts?

nikostoulas: I would rather not.  However, apps that would need initialize can call diamorphosis outside orka. Tested this in cerberus so I will drop this pr.

PR #19, opened 2019-05-24
Convert builder to class

### LINKED ISSUE #17: Make forTypescript work
@klesgidis you know how

klesgidis:
Make types to work for builder by converting it to class
Fixes #17

### DIFF
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -17,9 +17,114 @@ import assert = require('assert');
 import { Server } from 'http';
 import logo from './initializers/logo';
 
+export class OrkaBuilder {
+  options: Partial<OrkaOptions>;
+  config: any;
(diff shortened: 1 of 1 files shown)


src/builder.ts
```diff
@@ -17,9 +17,114 @@ import assert = require('assert');
 import { Server } from 'http';
 import logo from './initializers/logo';
 
+export class OrkaBuilder {
```
nikostoulas: 💣

nikostoulas: Does this need to be exported?

klesgidis: It complains here: 
https://github.com/Workable/orka/blob/master/src/index.ts#L4
and here:
https://github.com/Workable/orka/blob/master/src/orka.ts#L6

if i don't export it

PR #26, opened 2019-06-07
use lodash instead of is array

alkiskal:


nikostoulas:
Since the same pattern is used elsewhere go ahead and merge it for consistency

alkiskal:
Correct, I did it for the consistency, no other reason.

PR #30, opened 2019-06-13
Kafka

### LINKED ISSUE #18: Support kafka initialization

alkiskal:
Fixes #18 
Adds kafka support, example: https://github.com/Workable/maestro/pull/266

Next step: Make kafka dependencies optional through codependence.


examples/kafka-example/routes.js
```diff
@@ -0,0 +1,28 @@
...
+  brokers: config.kafka.brokers
+};
+const kafka = new Kafka(options);
+
+module.exports = {
+  get: {
+    '/read': async (ctx, next) => {
+      await kafka.connect();
+      const topic = config.kafka.consumer.topics.test;
+      new KafkaHandler(kafka, { topic });
```
nikostoulas: 👍

src/initializers/kafka/base-kafka-handler.ts
```diff
@@ -0,0 +1,43 @@
...
+  constructor(kafka: Kafka, options: { topic: string; logger: Logger; batchSize?: number; }) {
+    const { topic, batchSize, logger } = options;
+    this.topic = topic;
+    this.batchSize = batchSize || 1;
+    this.logger = logger;
+    this.consumer = kafka.createConsumer(this.topic);
+    this.consumer.connect().then(() => this.consume().catch(err => this.logger.error(err)));
+  }
+
+  async abstract handle(message: any): Promise<any>;
```
nikostoulas: Can you add typings in message? Or is it only the value?

alkiskal: I'll have to check. For now I need only the ok for the interface

nikostoulas: The example is clear. I like it

alkiskal: This is only the value, `await Promise.all(flatten([messages]).map(m => this.handle(m.value)));`
Moreover, this is already an object that we can use 
```
      false,
      true, //Receive as object
```

alkiskal: I believe we can close this for now. Maestro has a working example, we can come back for any improvements if the need arises.

src/initializers/kafka/base-kafka-handler.ts
```diff
@@ -0,0 +1,43 @@
...
+  batchSize: number;
+  logger: Logger;
+
+  constructor(kafka: Kafka, options: { topic: string; logger: Logger; batchSize?: number; }) {
+    const { topic, batchSize, logger } = options;
+    this.topic = topic;
+    this.batchSize = batchSize || 1;
+    this.logger = logger;
+    this.consumer = kafka.createConsumer(this.topic);
+    this.consumer.connect().then(() => this.consume().catch(err => this.logger.error(err)));
```
anastasios-workable: If this fails the instance is invalid. Can you move object instantiation to an async factory method instead?

src/initializers/kafka/kafka.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  public createConsumer(topic: string) {
+    const { groupId, brokers } = this.options;
+    const config = {
+      groupId,
+      logger: getLogger('orka.kafka.consumer.internal'),
+      noptions: {
+        'group.id': groupId,
+        'metadata.broker.list': brokers.join(','),
+        'socket.keepalive.enable': true,
```
anastasios-workable: `Socket keepalives need to be enabled on both sides to expedite connection termination on connectivity loss (e.g., network partitioning) for both parties. This means the broker needs to implement it too.` Does it?

src/initializers/kafka/auth-options.ts
```diff
@@ -0,0 +1,32 @@
...
+  };
+  logger.trace('Certificates created: ', paths);
+  return paths;
+};
+
+export default ({ key, cert, ca }) => {
+  const paths = certificates({ key, cert, ca });
+
+  return {
+    'security.protocol': <'ssl'> 'ssl',
```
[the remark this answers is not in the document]
alkiskal: It's a workaround for enums, so you can use string.

src/initializers/kafka/base-kafka-handler.ts
alkiskal: I can remove the `consumer.connect()` from the constructor and add it in the `consume`. Then the caller must call `consume()` explicitly. Does that work?

src/initializers/kafka/kafka.ts
alkiskal: I can't find anything related in lenses. I found in the issues section that it doesn't cause any harm if it's enabled. We can check further if you want, but I'll need assistance regarding the lenses/kafka configuration.

alkiskal:
@klesgidis It does not. https://github.com/nodefluent/node-sinek#you-might-also-like . But we don't need it. In librdkafka we used streams because it was easier to use and it handled reconnects etc, but in this it handles all that already. Plus the native consumer/producer are faster.

src/initializers/kafka/kafka.ts
anastasios-workable: Thanks for checking.

src/initializers/kafka/base-kafka-handler.ts
nikostoulas: What if the connection fails after a while? Can we handle reconnections somehow?

alkiskal: Supposedly it handles reconnections automatically
```
      this.consumer.on("disconnected", () => {
        if(this._inClosing){
          this._reset();
        }
        logger.warn("Disconnected.");
        //auto-reconnect --> handled by consumer.consume();
      });
```

I'll have to test locally (terminate the broker and bring it back up or something equivalent)

alkiskal: Used this https://hub.docker.com/r/landoop/kafka-lenses-dev/ locally in order to test what happens when terminating the broker. 
Everything works as expected. Both the consuming and producing resume automatically.

nikostoulas:
Can we start adding some unit tests?

PR #32, opened 2019-06-18
add  @types/koa-bodyparser

ilianakyl:
added @types/koa-bodyparser in order to use contect.request.body

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,6 @@
     "instrument": true
   },
   "dependencies": {
-    "@types/koa": "^2.0.48",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -43,6 +43,7 @@
   },
   "dependencies": {
     "@types/koa": "^2.0.48",
+    "@types/koa-bodyparser": "^4.3.0",
```
anastasios-workable: All @types should be `devDependencies` instead, since they are not really needed at runtime.

ilianakyl: Done

PR #33, opened 2019-06-21
export newrelic

alkiskal:


src/initializers/newrelic/index.ts
```diff
@@ -1,9 +1,12 @@
...
 export default async (config, orkaOptions: Partial<OrkaOptions>) => {
   if (config.newRelicLicenseKey) {
     process.env.NEW_RELIC_HOME = __dirname;
     process.env.NEW_RELIC_APP_NAME = `${orkaOptions.appName} ${config.nodeEnv}`;
-    await import('newrelic');
+    newrelic = await import('newrelic');
   }
 };
+
+export { newrelic };
```
nikostoulas: does this work? Have you tested it?

alkiskal: Tested with console logs and it gets updated.

PR #40, opened 2019-07-02
kafka improvements, sinek is optional

alkiskal:
- Makes kafka optional through codependency
- Connects to kafka inside orka
- Add more typings


examples/kafka-example/config.js
```diff
@@ -4,7 +4,7 @@ module.exports = {
     appName: 'foo'
   },
   kafka: {
-    brokers: [],
+    brokers: ["localhost:9092"],
```
nikostoulas: 👍

test/initializers/kafka/base-kafka-handler.test.ts
```diff
@@ -29,12 +39,12 @@ describe('base kafka handler class', async () => {
...
     sandbox.stub(Kafka.prototype, 'createConsumer').returns(consumeStub);
     const kafka = new Kafka({ groupId: 'groupId', clientId: 'clientId', brokers: [] } as any);
-    const handler = new TestKafkaHandler(kafka, { topic: 'topic', logger: getLogger('test'), batchSize: 2 });
+    const handler = new TestKafkaHandler(kafka, { topic: 'topic', logger, batchSize: 2 });
     await new Promise(resolve => setTimeout(resolve, 10));
     consumeStub.connect.calledOnce.should.eql(true);
     consumeStub.consume.calledOnce.should.eql(true);
     consumeStub.commit.calledOnce.should.eql(true);
-    handleStub.calledOnce.should.eql(true);
+    handleStub.calledTwice.should.eql(true);
```
nikostoulas: Why did this change?

alkiskal: Maybe this should be written a little bit better. `handleStub` is called once for `handle` and once for `handleSuccess` (just before commiting we call a callback which in our case sends the output to another topic)

nikostoulas: What's the use of handleSuccess?

PR #45, opened 2019-07-05
Proposal of how rabbitMQ handlers initialisation could be improved

nikostoulas:
@ilianakyl , @alkiskal wdyt?

Haven't tested it. If we agree with the signature @ilianakyl can you finish this on Monday?

### DIFF
--- a/examples/rabbitmq-example/app.js
+++ b/examples/rabbitmq-example/app.js
@@ -12,6 +12,6 @@ orka({
   diamorphosis: { configFolder: './examples/rabbitmq-example' },
   routesPath: './examples/rabbitmq-example/routes.js',
   logoPath: './examples/simple-example/logo.txt',
-  rabbitHandlers: () => new ExampleHandler('example_queue'),
+  rabbitHandlers: [() => new ExampleHandler('example_queue')],
   beforeStart: () => console.log(`Going to start env: ${config.nodeEnv}`)
 }).start();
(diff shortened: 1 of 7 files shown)


ilianakyl:
Sure! we can give it a try.

alkiskal:
Another approach would be to let the handlers handle the reconnect 
https://github.com/Workable/rabbit-queue/blob/master/ts/baseQueueHandler.ts
Since the Rabbit instance is available they could attach to the connected event in the constructor.

nikostoulas:
We could do this @alkiskal but this possibly requires a lot more work in rabbit-queue (possibly also to distinguish initial connect from reconnecting) . Totally worth it though. We should probably do a lot more work there to generalize stuff (also check the issues).
Care to do it?

Until then shall we merge this?

alkiskal:
- Yes I can do it.
- We could merge this, but when every handler will handle the connect/reconnect on itself wouldn't this require changes again in orka (source, typings) and the clients using it? Wouldn't make more sense to do something like this
```
if (!Rabbit.connecting) {
  init();
}
Rabbit.on('connected', init);
```
and when the change is ready in rabbit-queue we could just do `init();`

nikostoulas:
Changes in orka only. Clients could keep giving the init function in orka to call it.

alkiskal:
The init function instead of the handlers sounds better. We have bindToTopic calls as well

PR #47, opened 2019-07-05
use manual batching, add typings

alkiskal:


src/typings/kafka.d.ts
```diff
@@ -1,3 +1,16 @@
+export interface KafkaConfig {
+  kafka: {
```
nikostoulas: Why use this top-level key?

PR #56, opened 2019-07-27
Adds #callback to builder to return Koa request handler

echetzakis:
In `VHS` we use `supertest` for integration testing our api like so:
```
const request = require('supertest');
const app = require('../app');

global.route = request(app.callback());
```

Orka hides the Koa app instance hence we no longer have access to the Koa#callback
This PR adds a `callback` method to the builder which does exactly that.

### DIFF
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -115,6 +115,22 @@ export class OrkaBuilder {
     });
   }
 
+  // Return a request handler callback instead of starting the server
+  // Usefull for testing
+  async callback() {
+    const _logger = getLogger('orka');
+    try {
+      while (this.queue.length) {
+        await this.queue.shift()();
+      }
+      const koa = require('./initializers/koa');
+      return koa.callback(this.middlewares);
+    } catch (e) {
+      _logger.error(e);
+      process.exit(1);
+    }
+  }
+
   async start(port: number = this.config.port) {
     const _logger = getLogger('orka');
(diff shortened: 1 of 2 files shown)


src/builder.ts
```diff
@@ -21,152 +21,165 @@ import logo from './initializers/logo';
...
+            return router(routes)(...args);
+        });
+    }
+
+    // Return a request handler callback instead of starting the server
+    // Usefull for testing
+    callback() {
+        const _logger = getLogger('orka');
+        try {
+            const koa = require('./initializers/koa');
```
nikostoulas: Shouldn't you also add the 
```
while (this.queue.length) {
                await this.queue.shift()();
 }
```

echetzakis: 👍

nikostoulas:
Nice @echetzakis. Do you have an npm account to add you in the organization?

echetzakis:
> Nice @echetzakis
> Can you make sure before merging that src/builder.ts is formatted using prettier?

👍 

> 
> Note that you can test this directly by using:
> "@workablehr/orka": "git://github.com/Workable/orka.git#feature/koa_callback"
> in your package json.

Already doing so [in my VHS feature](https://github.com/Workable/vhs/blob/78acfca247909c602f381a589b4e851a9c9f729a/package.json#L59)

PR #58, opened 2019-07-30
Add init for nc usage

alkiskal:


src/builder.ts
```diff
@@ -131,7 +131,7 @@ export class OrkaBuilder {
     }
   }
 
-  async start(port: number = this.config.port) {
+  async init() {
```
nikostoulas:
```suggestion
  async initQueues() {
``` 
?

src/builder.ts
```diff
@@ -140,6 +140,17 @@ export class OrkaBuilder {
...
+      _logger.error(e);
+      process.exit(1);
+    }
+    return this;
+  }
+
+  async start(port: number = this.config.port) {
+    const _logger = getLogger('orka');
+    try {
+      await this.init();
```
nikostoulas: I would move the logging here not in init. Since it also mentions middlewares.

src/builder.ts
nikostoulas: Use it also in a callback. Maybe the try-catch is not needed here.

PR #60, opened 2019-07-31
Allow honeybadger env name to be different than NODE_ENV

echetzakis:


nikostoulas:
@echetzakis out of curiosity what is the case that you want the honeybadger env to be different?

PR #61, opened 2019-08-01
Bump rabbit and export it.

nikostoulas:
bump_rabbit


alkiskal:
Maybe make it optional as well? So the client can decide what version to use.

PR #63, opened 2019-08-01
Expose producer send arguments

klesgidis:


src/initializers/kafka/kafka.ts
```diff
@@ -25,8 +25,8 @@ export default class Kafka {
     logger.info(`Kafka connected ${brokers.join(',')}`);
   }
 
-  public async send(topic: string, message: string | Buffer) {
-    const { key, partition, offset } = await this.producer.send(topic, message);
+  public async send(topic: string, message: string | Buffer, _partition?: number, _key?: string) {
```
alkiskal: nit: I believe the underscore is used as a convention in sinek for optional arguments. We are using typescript so there is no need for a convention like that. You can remove the underscore from the names.

src/initializers/kafka/index.ts
```diff
@@ -1,9 +1,10 @@
 import { isEmpty } from 'lodash';
 import { KafkaConfig } from '../../typings/kafka';
+import Kafka from './kafka';
 
 export { default as BaseKafkaHandler } from './base-kafka-handler';
 
-let kafka;
+let kafka: Kafka;
```
alkiskal: Nice!

PR #68, opened 2019-08-05
Adds support for `credentials = true` when setting up the cors properties.

anastasios-workable:


src/builder.ts
```diff
@@ -52,11 +52,13 @@ export class OrkaBuilder {
     return this;
   }
 
-  useCors(allowedOrigins = this.config.allowedOrigins) {
+  useCors({ credentials = undefined, allowedOrigins = this.config.allowedOrigins } = this.config.cors || {}) {
```
klesgidis: 👍

PR #70, opened 2019-08-06
Always initialize process.env.NEW_RELIC_HOME to point to orka, otherwise if the root project is missing a newrelic.js config file, it is not configured properly.

anastasios-workable:


src/initializers/newrelic/index.ts
```diff
@@ -3,8 +3,8 @@ import requireInjected from '../../require-injected';
 
 let newrelic;
 export default async (config, orkaOptions: Partial<OrkaOptions>) => {
+  process.env.NEW_RELIC_HOME = __dirname;
```
nikostoulas: Is this to be used for development environments?

anastasios-workable: No.

nikostoulas: 👍

PR #73, opened 2019-08-07
Fix newrelic initialization

nikostoulas:
fix_newrelic

### DIFF
--- a/examples/builder-example/app.js
+++ b/examples/builder-example/app.js
@@ -3,7 +3,7 @@ const { builder } = require('../../build');
 const staticOptions = { diamorphosis: { configFolder: './examples/simple-example' } };
 
 const w = builder(staticOptions)
-  .forTypescript()
+  .forTypescript(false)
   .useDefaults()
   .with(config => console.log(`Going to Start env: ${config.nodeEnv}`))
   .use(async (ctx, next) => {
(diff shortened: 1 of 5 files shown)


src/builder.ts
```diff
@@ -9,7 +9,7 @@ export default (defaults: Partial<OrkaOptions> = _defaults) => {
   const options: Partial<OrkaOptions> = lodash.cloneDeep(lodash.defaultsDeep({}, defaults, _defaults));
 
   // Always call newrelic
-  newrelic(this.config, defaults);
+  newrelic(defaults.appName || process.env.APP_NAME);
```
anastasios-workable:
```suggestion
  newrelic(defaults.appName || (config.app && config.app.name) || process.env.APP_NAME);
```

examples/builder-example/app.js
```diff
@@ -3,7 +3,7 @@ const { builder } = require('../../build');
 const staticOptions = { diamorphosis: { configFolder: './examples/simple-example' } };
 
 const w = builder(staticOptions)
-  .forTypescript()
+  .forTypescript(false)
```
klesgidis: why?

nikostoulas: Because the example is in js not in ts. 
But you get how you could use it.
Feel free to add ts examples!

PR #80, opened 2019-09-12
Add workaround for middlewares that need access to config

alkiskal:
WIP: This is not ready yet

### DIFF
--- a/src/orka-builder.ts
+++ b/src/orka-builder.ts
@@ -12,7 +12,7 @@ import { getLogger } from './initializers/log4js';
 import riviere from './initializers/koa/riviere';
 import addRequestId from './initializers/koa/add-request-id';
 import _defaults from './default-options';
-import { OrkaOptions } from './typings/orka';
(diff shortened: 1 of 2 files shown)


src/orka-builder.ts
```diff
@@ -34,9 +34,9 @@ export default class OrkaBuilder {
     this.queue = [];
   }
 
-  use(m: Middleware<any> | Middleware<any>[] = []) {
+  use(m: ReturnsMiddleware<any> | ReturnsMiddleware<any>[] | Middleware<any> | Middleware<any>[] = []) {
     m = lodash.flatten([m]).filter(lodash.identity);
-    m.forEach(__ => this.middlewares.push(__));
+    m.forEach(c => this.middlewares.push(c.arguments === 1 ? (c as ReturnsMiddleware<any>)(this.config) : c));
```
nikostoulas: This isn't a bad idea.  `c.length === "number of arguments"`

alkiskal: `c.arguments.length` correct.

PR #82, opened 2019-09-17
Two steps init option

alkiskal:


examples/simple-example-two-steps/logo.txt
```diff
@@ -0,0 +1,15 @@
+|
```
nikostoulas: rm use it from the other example

test/examples/examples.test.ts
```diff
@@ -7,7 +7,12 @@ const ws: [string, string, Function?][] = [
   ['../../examples/simple-example/app', 'simple-example', () => delete process.env.NEW_RELIC_LICENSE_KEY],
   ['../../examples/builder-example/app', 'builder-example', () => delete process.env.NEW_RELIC_LICENSE_KEY],
   ['../../examples/simple-example/app', 'simple-example newrelic', () => (process.env.NEW_RELIC_LICENSE_KEY = 'foo')],
-  ['../../examples/builder-example/app', 'builder-example newrelic', () => (process.env.NEW_RELIC_LICENSE_KEY = 'foo')]
+  ['../../examples/builder-example/app', 'builder-example newrelic', () => (process.env.NEW_RELIC_LICENSE_KEY = 'foo')],
+  [
+    '../../examples/simple-example-two-steps/app',
+    'simple-example-two-steps newrelic',
+    () => process.env.NEW_RELIC_LICENSE_KEY
```
nikostoulas: why do you return newrelic here?

examples/simple-example-two-steps/config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
nikostoulas: rm use from other example

examples/simple-example-two-steps/routes.js
```diff
@@ -0,0 +1,20 @@
+module.exports = {
```
nikostoulas: You could also use the routes from simple example

src/orka.ts
```diff
@@ -2,9 +2,10 @@ import * as lodash from 'lodash';
 import builder from './builder';
 import defaults from './default-options';
 import { OrkaOptions } from './typings/orka';
+import OrkaBuilder from './orka-builder';
 
-const fromOptions = (options: Partial<OrkaOptions>) => {
-  return builder(lodash.defaultsDeep(options, defaults))
+const fromOptions = (options: Partial<OrkaOptions & { builder?: OrkaBuilder }>) => {
```
nikostoulas: Why don't you update the OrkaOptions?

src/orka.ts
```diff
@@ -2,9 +2,10 @@ import * as lodash from 'lodash';
...
+  return (options.builder || builder(lodash.defaultsDeep(options, defaults)))
```
nikostoulas: Nice!

src/typings/orka.d.ts
```diff
@@ -1,2 +1,4 @@
 import options from '../default-options';
-declare type OrkaOptions = typeof options;
+import OrkaBuilder from '../orka-builder';
+
+declare type OrkaOptions = typeof options & { builder?: OrkaBuilder };
```
nikostoulas: not here but in default-options

src/orka.ts
nikostoulas: What about the new configs do they pass again?

src/default-options.ts
```diff
@@ -26,7 +27,8 @@ export default {
     },
     groupId: '',
     clientId: '',
-    brokers: []
+    brokers: [],
+    orkaBuilder: null as OrkaBuilder
```
nikostoulas:
```suggestion
    builder: null as OrkaBuilder
```

PR #85, opened 2019-09-19
Initialize middleware after tasks

nikostoulas:
This is a breaking change.
Instead of using
```js
builder.use((ctx, next) => ctx.body = 'ok')
```
we now will have:
```js
builder.use(()=> (ctx, next) => ctx.body = 'ok')
```

This is essential so that middlewares are initialized after tasks.
A good example is routes. In most apps, we need routes after some initializations have been made inside tasks. 
In the latest tag to overcome this limitation routes were initialized in a callback. That means routes were parsed in every HTTP request which is very very wrong.


src/initializers/koa/add-request-id.ts
```diff
@@ -1,7 +1,7 @@
 import { Context, Middleware } from 'koa';
 
 export default function(config): Middleware {
-  return async function(ctx: Context, next: () => void) {
+  return async function addRequestId(ctx: Context, next: () => void) {
```
klesgidis: Is this added for logging middleware name to newrelic?

src/initializers/koa/index.ts
```diff
@@ -1,14 +1,16 @@
...
+export const listen = async (app: Koa<any, {}>, port: number, middlewares: Middleware<any>[], callback: () => any) => {
   middlewares.forEach(middleware => app.use(middleware));
   return app.listen(port, callback);
 };
 
-export const callback = (middlewares: Middleware<any>[]) => {
-    let app = new Koa();
-    middlewares.forEach(middleware => app.use(middleware));
-    return app.callback();
+export const callback = (app: Koa<any, {}>, middlewares: Middleware<any>[]) => {
```
klesgidis:
```suggestion
export const callback = (app: Koa<any, {}>, middlewares: Middleware<any>[] = []) => {
```

src/initializers/koa/add-request-id.ts
nikostoulas: Yes, it would be nice if newrelic could log these

PR #89, opened 2019-10-02
Implement json appender for log4js

manoskouvarakis:


src/initializers/log4js/json-appender.ts
```diff
@@ -0,0 +1,42 @@
+import * as lodash from 'lodash';
+
+const jsonAppender = ({ filter_status = [] }) => {
```
nikostoulas: Do we use the filter_status? It is used in honeybadger appender to not send statuses that are expected eg: 401 unauthorized.

src/initializers/log4js/json-appender.ts
```diff
@@ -0,0 +1,42 @@
...
+  return logEvent => {
+    let event = logEvent.level.levelStr === 'ERROR' ? createErrorLog(logEvent) : createValidLog(logEvent);
+    var json = JSON.stringify(event);
+    console.log(json);
```
nikostoulas: What does log4js use internally? Maybe process.stdout.write is more suitable

src/initializers/log4js/json-appender.ts
```diff
@@ -0,0 +1,42 @@
...
+    message: logEvent.data[0].message,
+    stack_trace: logEvent.data[0].stack
+  };
+};
+
+const createValidLog = logEvent => {
+  const data = lodash.flattenDeep(logEvent.data);
+  const message = data
+    .map(element => {
+      return typeof element === 'object' ? JSON.stringify(element) : element;
```
nikostoulas: Can we use append the object keys to the returned obj?

src/initializers/log4js/json-appender.ts
manoskouvarakis: Done

src/initializers/log4js/json-appender.ts
manoskouvarakis: Done (kept console as it has only memory overhead)

src/initializers/log4js/json-appender.ts
manoskouvarakis: Done

src/initializers/log4js/json-appender.ts
```diff
@@ -0,0 +1,54 @@
...
+  };
+};
+
+const createValidLog = logEvent => {
+  const data = lodash.flattenDeep(logEvent.data);
+  let context = {};
+  const message = data
+    .map(element => {
+      if (typeof element === 'object') {
+        Object.assign(context, element);
```
nikostoulas: It's a bit weird that a map has side effects. Can we refactor it to use forEach instead?

src/initializers/log4js/json-appender.ts
nikostoulas: 👍

src/initializers/diamorphosis.ts
```diff
@@ -5,6 +5,7 @@ export default (config, orkaOptions: Partial<OrkaOptions>) => {
   config.nodeEnv = config.nodeEnv || 'development';
   config.honeybadgerApiKey = config.honeybadgerApiKey || '';
   config.honeybadgerEnvironment = config.honeybadgerEnvironment || '';
+  config.honeybadgerFilterStatus = config.honeybadgerFilterStatus || [];
```
nikostoulas: 👍

src/initializers/log4js/json-appender.ts
manoskouvarakis: Done

src/initializers/diamorphosis.ts
```diff
@@ -23,5 +30,4 @@ export default (config, orkaOptions: Partial<OrkaOptions>) => {
     ...config.riviere
   };
   diamorphosis(orkaOptions.diamorphosis);
-  config.honeybadgerEnvironment = config.honeybadgerEnvironment || config.nodeEnv;
```
nikostoulas: Can you add this here again? And possibly add a test for it.
The problem is that if you don't set honeybadgerEnvironment but only set NODE_EVN in env the honeybadgerEnvironment remains 'development'

src/initializers/diamorphosis.ts
```diff
@@ -1,14 +1,21 @@
...
+    apiKey: '',
+    environment: config.nodeEnv,
+    filterStatus: '',
+    ...config.honeybadger
+  };
+  config.printLogo = defaultTo(config.printLogo, true);
   config.log = {
     pattern: '%[[%d] [%p] %c%] %m',
     level: 'debug',
+    consoleLogging: true,
```
nikostoulas: LOG_CONSOLE, LOG_JSON seems better wdyt?

(So log: {console: true, json:false})

PR #91, opened 2019-10-10
Simplify default configuration for json logging

manoskouvarakis:


src/initializers/diamorphosis.ts
```diff
@@ -25,10 +25,18 @@ export default (config, orkaOptions: Partial<OrkaOptions>) => {
...
+    styles: [],
     headersRegex: '^X-.*',
     ...config.riviere
   };
   diamorphosis(orkaOptions.diamorphosis);
   config.honeybadger.environment = config.honeybadger.environment || config.nodeEnv;
+
+  if (config.log.console === '') {
+    config.log.console = !config.log.json;
+  }
```
nikostoulas: if it is not empty can we make it Boolean?

test/diamorphosis/diamorphosis.test.ts
```diff
@@ -59,4 +59,97 @@ describe('Diamorphosis Test', () => {
...
+          envFolder: path.resolve('./test/diamorphosis/env')
+        }
+      } as OrkaOptions;
+    });
+
+    afterEach(() => {
+      delete require.cache[require.resolve(options.diamorphosis.configPath)];
+    });
+
+    it('nothing set => console:true, json:false, styles:[] ', () => {
```
nikostoulas: you could use context('when nothing is set config', ()=> it('should be console:true, json:false, styles:[]})

test/diamorphosis/diamorphosis.test.ts
```diff
@@ -59,4 +59,97 @@ describe('Diamorphosis Test', () => {
...
+
+      diamorphosis(config, {} as OrkaOptions);
+
+      config.log.console.should.equal(false);
+      config.log.json.should.equal(true);
+      config.riviere.styles.length.should.equal(1);
+      config.riviere.styles[0].should.equal('json');
+    });
+
+    it('console:true, json:true, styles:[] => console:true, json:true, styles:["json"] ', () => {
```
nikostoulas: Can you add a test where instead of providing the config you provide env variables in process.env?
This way you can test the case I mention above

src/initializers/diamorphosis.ts
manoskouvarakis: Done

test/diamorphosis/diamorphosis.test.ts
manoskouvarakis: Done

test/diamorphosis/diamorphosis.test.ts
manoskouvarakis: Done

PR #100, opened 2019-10-19
Proposal: Adds custom trace id to each log

klesgidis:
## Summary
Appends a custom trace id to each log context. You can specify which is the header you want to append to logs in your koa middlewares.

@nikostoulas worth a try, what you think?

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,8 @@
   },
   "dependencies": {
     "@workablehr/riviere": "^1.4.0",
+    "cls-hooked": "^4.2.2",
+    "cls-rtracer": "^1.3.0",
     "codependency": "^2.1.0",
     "diamorphosis": "^0.8.1",
     "fast-koa-router": "^1.0.1",
--- a/src/initializers/log4js/json-appender.ts
+++ b/src/initializers/log4js/json-appender.ts
@@ -1,4 +1,5 @@
 import * as lodash from 'lodash';
+import * as rTracer from 'cls-rtracer';
 
 const jsonAppender = () => {
   return logEvent => {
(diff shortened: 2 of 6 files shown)


src/orka-builder.ts
```diff
@@ -79,6 +80,21 @@ export default class OrkaBuilder {
     );
   }
 
+  // header name defaults to X-Request-Id case insensitive
```
nikostoulas: we already have a config.traceHeaderName can we use this instead?

src/initializers/log4js/json-appender.ts
```diff
@@ -24,9 +31,15 @@ export const createErrorLog = logEvent => {
 };
 
 export const createValidLog = logEvent => {
-  const data = lodash.flattenDeep(logEvent.data);
+  const { context: _context = {}, data: _data } = logEvent;
+  const data = lodash.flattenDeep(_data);
+  const trace = rTracer.id();
```
nikostoulas: Can we do this for all appenders and not only for json?

manoskouvarakis:
Depending on the flexibility we need (if we want something too sophisticated this approach may not be appropriate), we can also investigate the `addContext()` method that log4js supports. 
https://github.com/log4js-node/log4js-node/blob/a3f494f87f915324bc21430cef310e48a3a5fa26/docs/api.md

In order to add a simple key-value pair in all messages you need to:
- Configure layout to support it `pattern: '%d %p %c %X{contextVariable}'`
- Use addContext method somewhere globally `logger.addContext('contextVariable', 'bla')`
- Enhance jsonLogger to support it

package.json
```diff
@@ -53,6 +53,8 @@
   },
   "dependencies": {
     "@workablehr/riviere": "^1.4.0",
+    "cls-hooked": "^4.2.2",
+    "cls-rtracer": "^1.3.0",
```
nikostoulas: Do we need cls-rtracer? We already handle requestId and this seems like sth easy to implement with cls-hooked only. @klesgidis wdyt?

klesgidis: I saw the code that handles the context and the namespaces and I think it's more easier to just use the middleware and let them do the work for us. You can have a look here: https://github.com/puzpuzpuz/cls-rtracer/blob/master/index.js#L47

PR #101, opened 2019-10-23
Fix json-appender to handle logs with parameters

manoskouvarakis:


src/initializers/log4js/json-appender.ts
```diff
@@ -48,10 +48,13 @@ const getContextObject = data => {
...
-export function configure() {
-  return jsonAppender();
+function configure(config, layouts) {
+  let layout = layouts.messagePassThroughLayout;
+  return jsonAppender(layout);
 }
+
+module.exports.configure = configure;
+
+// export function configure() {
```
nikosd23: Why is this commented out?

src/initializers/log4js/json-appender.ts
```diff
@@ -23,11 +23,11 @@ export const createErrorLog = logEvent => {
   };
 };
 
-export const createValidLog = logEvent => {
+export const createValidLog = (layout, logEvent) => {
```
nikosd23: What if you don't pass layout?

src/initializers/log4js/json-appender.ts
manoskouvarakis: Fixed

src/initializers/log4js/json-appender.ts
manoskouvarakis: This method is exported only for testing purposes. So layout will be passed from json appender initialisation

test/env.ts
```diff
@@ -2,3 +2,4 @@ process.env.LOG_LEVEL = 'fatal';
 process.env.RIVIERE = 'false';
```
manoskouvarakis: This line should be removed with the new enabled property

nikostoulas: Done

PR #108, opened 2019-11-04
Use app.env in newrelic and honeybadger

### LINKED ISSUE #107: Override Newrelic application name
Currently, the application name that we report to new relic is based on the `NODE_ENV` env variable. Apps that set this var to `production` in staging as well have no way to report in different account per environment (stg1, stg2 etc)

Let's change the rule to consult a `NEW_RELIC_APP_NAME` env var before

nikostoulas:
Fixes #107

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,13 @@ Orka uses some defaults for the below values in your config.js
 ```js
 config = {
   nodeEnv :'development';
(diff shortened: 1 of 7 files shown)


nikostoulas:
@echetzakis @kzacharakis if you use HONEYBADGER_ENV you will have to rename it to APP_ENV after this!

PR #110, opened 2019-11-07
Create custom replacer for circular json

manoskouvarakis:
A custom replacer is implemented to support circular json. 

When the replacer finds a circular structure it prints "circular_ref" as a value to the specific field


examples/simple-example/routes.js
```diff
@@ -11,6 +11,27 @@ module.exports = {
     '/logError': async (ctx, next) => {
       getLogger('log').error(new Error('test'), 'this was a test error', { context: 'foo' });
       ctx.throw(new Error('test'), 505);
+    },
+    '/logCircular': async (ctx, next) => {
```
nikostoulas: Doesn't seem simple example 😛

src/initializers/log4js/json-appender.ts
```diff
@@ -4,11 +4,32 @@ const jsonAppender = layout => {
...
   };
 };
 
+const safeStringify = event => {
+  try {
+    return JSON.stringify(event);
+  } catch (error) {
+    let seen = new WeakSet();
+
+    const circularJson = JSON.stringify(event, (key, value) => {
```
nikostoulas: Did you try util.inspect?

manoskouvarakis: Yes.
The reason I preferred a custom implementation is that utils.inspect documentation says that is intended for debugging purposes and not depend on it programmatically.
http://nodejs.org/api/util.html#util_util_inspect_object_options

alkiskal:
@manoskouvarakis semaphore is failing. please check. 
Moreover, great job in setting weakset to null at the end ensuring that it will be garbage collected.
You could have this method somewhere else (helper or sth) so it can be used in other places as well if applicable. Other than that I believe it's good to go

src/initializers/log4js/json-appender.ts
nikostoulas: What is log4js using internally? What would it do if you passed it a circular object?

manoskouvarakis: It uses https://www.npmjs.com/package/flatted which is a library to handle circular dependencies. This is the commit:
https://github.com/log4js-node/log4js-node/commit/29c92211f2b4f030186392f000501d107ae4fd3d

PR #114, opened 2019-11-15
Bump sinek to 8.1.0 (uses node-rdkafka 2.7.4) + bump minors

klesgidis:
## Summary
* Bump sinek to 8.1.0 that uses node-rdkafka 2.7.4. Fixes the issue on running kafka on macOS
* Bump minor versions


nikostoulas:
@klesgidis note tha sinek is optional in orka so you should bump the version in your repo too. Actually bumping the version in your repo should suffice.

PR #116, opened 2019-11-20
Teach orka redis

nikostoulas:
Redis initialization through orka similar to how mongodb works.
Because redis initialization is more difficult we export 2 methods getRedis to get the client created by orka and createRedisConnection to create extra connections.

To enable ssl you need to set 3 env variables:

```
REDIS_OPTIONS_TLS_CA
REDIS_OPTIONS_TLS_CERT
REDIS_OPTIONS_TLS_KEY
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,16 @@ config = {
   app: {
     name: 'orka', // will be used as newrelic name, rabbit-queue prefix
     env: 'demo' // will be used as newrelic, honeybadger env - defaults to NODE_ENV
+  },
+  mongodb: {
+    url: '',
+    options: {}
+  },
+  redis: {
+    url:'',
+    options: {
+      
+    }
   }
   honeybadger:{
     apiKey: '', // will not add honeybadger by default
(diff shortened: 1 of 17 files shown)


src/initializers/redis.ts
```diff
@@ -0,0 +1,62 @@
...
+
+function getRedisUrl(config) {
+  return config && config.url;
+}
+
+function getHost(url) {
+  return url.split('@')[1] || url;
+}
+
+const exhastError = new Error('Retry retry_strategy options.total_retry_time exhausted');
```
alkiskal: `exhaust`?

README.md
```diff
@@ -29,6 +29,14 @@ config = {
   app: {
     name: 'orka', // will be used as newrelic name, rabbit-queue prefix
     env: 'demo' // will be used as newrelic, honeybadger env - defaults to NODE_ENV
+  },
```
klesgidis: 💯

src/initializers/redis.ts
nikostoulas: What do you mean @alkiskal ?

klesgidis: typo

nikostoulas: Thanks fixed

PR #120, opened 2019-11-29
Require redis after newrelic

nikostoulas:


src/initializers/redis.ts
```diff
@@ -1,4 +1,4 @@
-import { createClient, RedisClient } from 'redis';
+import { createClient as createClientType, RedisClient as RedisClientType } from 'redis';
```
anastasios-workable: Isn't that still importing the module?

nikostoulas: No. Typescript is clever enough to not include files or code only used for typings.

PR #130, opened 2020-02-16
add default routes

alkiskal:
Solves https://github.com/Workable/orka/issues/31

- A routes file has been added including health and stataus (more can be added if needed)
- The default routes can be overridden if the implementation for those routes is provided

### DIFF
--- a/src/initializers/koa/routes.ts
+++ b/src/initializers/koa/routes.ts
@@ -1,0 +1,23 @@
+import * as os from 'os';
+
+export default {
+  get: {
+    '/health': async ctx => (ctx.status = 200),
+    '/ping': async ctx => (ctx.status = 200),
+    '/status': async ctx => {
+      const mem = process.memoryUsage();
+      const m = 1024 * 1024;
+
+      ctx.body = {
+        rss: (mem.rss / m).toFixed(2) + 'M',
+        heapTotal: (mem.heapTotal / m).toFixed(2) + 'M',
+        heapUsed: (mem.heapUsed / m).toFixed(2) + 'M',
+        loadAvg: os.loadavg()
+      };
+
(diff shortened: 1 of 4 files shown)


src/initializers/koa/routes.ts
```diff
@@ -0,0 +1,23 @@
+import * as os from 'os';
+
+export default {
+  get: {
+    '/health': async ctx => (ctx.status = 200),
+    '/ping': async ctx => (ctx.status = 200),
```
nikostoulas: rm

src/initializers/koa/routes.ts
```diff
@@ -0,0 +1,23 @@
+import * as os from 'os';
+
+export default {
```
nikostoulas: I was thinking to just export the middleware. Not add them to the routes by the default. Don't have a strong opinion though. @anastasios-workable @nikosd23 wdyt?

src/initializers/koa/routes.ts
```diff
@@ -0,0 +1,23 @@
...
+  get: {
+    '/health': async ctx => (ctx.status = 200),
```
nikostoulas: Since orka knows if you use: kafka, redis, mongo or rabbit we could add those connections in the health endpoint. If the connections are not healthy we should not return 200.

alkiskal: I'm not opposed. I believe it's a good practice and I believe some teams are already doing this.
I'm troubled by the following though:
1. It's very intrusive and some teams might not want it (this can be solved by configuration though)
2. These checks are not "free", meaning that health checks should have a reasonable interval (check with sre team)

nikostoulas:
This was handled by #135 . Closing this.

Issue #129, opened 2020-02-16
Add joi validation related code


alkiskal:
We could include the following in orka validateParams and validation-error (check campaigns repo) in orka. We are using joi more and more in order to validate the http contacts and this code keeps being duplicated. This will effectively add @hapi/joi in orka so we should decide if we want to control the version from orka or allow that to the app (like newrelic)

PR #131, opened 2020-02-16
add validate params middleware

alkiskal:
Solves https://github.com/Workable/orka/issues/129
This functionality keeps being duplicated across sourcing apps.

- Add validate-params and expose methods
- Add @hapi/joi as a dependency (not injected, it's something that it makes sense to be controlled via orka)
- Refactored validation logic (taken from campaigns).
- Added TODO for validateQueryString (as is now, we only validate, but we do not store the cast result in  order to use it later on, all params remain as text)

One idea would be to add them in request.body as well, and depending on the order of the validation (for a request that contains both body and query params) the caller can set which param will remain (override logic). e.g.

POST localhost:3000/testPost?key=2
{
   "key": 3
}

depending on the validation we could either throw an assertion error, or just keep one of the two. For:
- validateQueryString(), validateBody() -> keep body (3)
- validateBody(), validateQueryString() -> keep query param (2)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -10,3 +10,4 @@ export { getKafka } from './initializers/kafka';
 export { getRabbit } from './initializers/rabbitmq';
 export { getRedis, createRedisConnection } from './initializers/redis';
 export { middlewares };
+export { validateBody, validateQueryString } from './initializers/koa/validate-params';
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -63,6 +63,14 @@
     {
       "type": "node",
       "request": "launch",
+      "name": "Validation example",
+      "program": "${workspaceFolder}/examples/validation-example/app.js",
+      "outFiles": ["${workspaceFolder}/**/*.js"],
+      "preLaunchTask": "build"
+    },
+    {
+      "type": "node",
+      "request": "launch",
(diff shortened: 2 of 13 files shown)


src/index.ts
```diff
@@ -10,3 +10,4 @@ export { getKafka } from './initializers/kafka';
 export { getRabbit } from './initializers/rabbitmq';
 export { getRedis, createRedisConnection } from './initializers/redis';
 export { middlewares };
+export { validateBody, validateQueryString } from './initializers/koa/validate-params';
```
nikostoulas: Could this be added in the middlewares folder? I think it matches with middlewares.

alkiskal: I'm not opposed to it, but what about riviere and add-request-id? Are they not considered middlewares as well?

alkiskal: Moreover, should I add the errors folder under middlewares as well or leave them in koa folder?

nikostoulas: Those 2 are not exposed to the client. They are internal.

nikostoulas: So let's leave them where they are for now.

alkiskal: Done

nikostoulas: Were they moved to middelewares?

PR #135, opened 2020-03-10
Adds health middleware that checks mongo connection health

nikosd23:


README.md
```diff
@@ -123,3 +123,22 @@ If you're going to use kafka from a MacOS, you'll need to add the following to y
...
 ```
+
+## Health Middleware
+It supports a health middleware that currently checks the availablity of mongo
+configuration and responds wheather the connectivity with the underlying database is stable. This is essential for liveness probes of several cloud providers such as K8s.
+
+Example route configuration
+
+```js
+const { health } = require('../../build/middlewares');
```
nikostoulas:
```suggestion
const { middlewares: {health} } = require('@workablehr/orka');
```

nikostoulas: or 
```suggestion
const { health } = require('@workablehr/orka/middlewares');
```
But then we could skip adding it in index.ts

examples/mongodb-example/routes.js
```diff
@@ -1,7 +1,9 @@
 const User = require('./user');
+const { health } = require('../../build/middlewares');
```
nikostoulas:
```suggestion
const { middlewares: {health} } = require('../../build');
``` 
Just to check that index.ts works.

src/middlewares/health.ts
```diff
@@ -0,0 +1,12 @@
+import { Context } from 'koa';
+import { getConnection } from '../initializers/mongodb';
+
+export default async function(ctx: Context) {
+  // tslint:disable-next-line: no-empty
+  const mongoConnection = getConnection(() => {});
+  if (!mongoConnection || mongoConnection.readyState === 1) {
```
klesgidis:
```suggestion
  if (!mongoConnection || mongoConnection.readyState === mongoose.ConnectionStates.connected) {
```

PR #142, opened 2020-04-14
Adds autoOffsetReset option for kafka consumers

echetzakis:
## With this PR: 
You can configure the `auto.offset.reset` property to `earliest` or `latest`.

Orka 's default remains `earliest` in order to be backwards compatible.
However [Kaflka's default seems to be](https://kafka.apache.org/documentation/#consumerconfigs) `latest`

### DIFF
--- a/src/initializers/kafka/kafka.ts
+++ b/src/initializers/kafka/kafka.ts
@@ -30,7 +30,7 @@ export default class Kafka {
     getLogger('orka.kafka.send').info(`partition(${_partition})[${offset}][${_key}] produced for topic ${topic}`);
   }
 
-  public createConsumer(topic: string) {
+  public createConsumer(topic: string, autoOffsetReset?: 'earliest' | 'latest') {
     const { groupId, brokers } = this.options;
     const config = {
       groupId,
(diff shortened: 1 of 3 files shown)


src/initializers/kafka/kafka.ts
```diff
@@ -42,10 +42,11 @@ export default class Kafka {
         'api.version.request': true,
         'socket.blocking.max.ms': 100,
         'log.connection.close': false,
+        'enable.auto.commit': false,
```
nikostoulas, line 45: Should this also be configurable? Is this a breaking change?

echetzakis: @nikostoulas I guess this has to do with the implementation of `base-kafka-handler`
https://github.com/Workable/orka/blob/7beb107871ba3748bc770b90e0c3cb370f75f9f4/src/initializers/kafka/base-kafka-handler.ts#L35
That calls explicitly `commit`, I guess @alkiskal can give us some more detail.

alkiskal: Yes, the whole implementation of the handler depends on this to be false.

echetzakis: oops! my bad, I was making some experiments, and forgot to remove it before opening the PR 😞

Issue #133, opened 2020-02-16
Kafka add header support


alkiskal:
https://github.com/Blizzard/node-rdkafka/commit/ba634bb71a134b794ae0a06b481af00e577cc2a4#diff-9d3a05693e055546138344d8145b175c

https://github.com/nodefluent/node-sinek/issues/121

alkiskal:
https://github.com/nodefluent/node-sinek/pull/151

This is now supported by node-sinek

PR #144, opened 2020-04-20
Add support for headers (node-sinek v9.1.0)

alkiskal:
- Solves #133
- Adds node-rdkafka in devDepencencies in order to run the example (node-sinek has it as optional now)

https://github.com/nodefluent/node-sinek/pull/151

### DIFF
--- a/src/initializers/kafka/index.ts
+++ b/src/initializers/kafka/index.ts
@@ -20,5 +20,3 @@ export const getKafka = () => {
   }
   return kafka;
 };
-
-export { Kafka };
--- a/src/initializers/kafka/index.ts
+++ b/src/initializers/kafka/index.ts
@@ -20,3 +20,5 @@ export const getKafka = () => {
   }
   return kafka;
 };
+
+export { Kafka };
--- a/src/initializers/kafka/kafka.ts
+++ b/src/initializers/kafka/kafka.ts
@@ -25,8 +25,21 @@ export default class Kafka {
     logger.info(`Kafka connected ${brokers.join(',')}`);
   }
 
-  public async send(topic: string, message: string | Buffer, key?: string, partition?: number) {
-    const { key: _key, partition: _partition, offset } = await this.producer.send(topic, message, partition, key);
+  public async send(
+    topic: string,
+    message: string | Buffer,
+    key?: string,
+    partition?: number,
+    headers?: { [key: string]: string }[]
+  ) {
+    const { key: _key, partition: _partition, offset } = await this.producer.send(
+      topic,
+      message,
+      partition,
+      key,
+      null,
+      headers
+    );
     getLogger('orka.kafka.send').info(`partition(${_partition})[${offset}][${_key}] produced for topic ${topic}`);
   }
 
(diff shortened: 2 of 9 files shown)


src/initializers/kafka/kafka.ts
```diff
@@ -25,8 +25,21 @@ export default class Kafka {
...
+    key?: string,
+    partition?: number,
+    headers?: { [key: string]: string }[]
+  ) {
+    const { key: _key, partition: _partition, offset } = await this.producer.send(
+      topic,
+      message,
+      partition,
+      key,
+      null,
```
nikosd23, line 40: what is this null here?

src/initializers/kafka/kafka.ts
```diff
@@ -25,8 +25,21 @@ export default class Kafka {
...
+      headers
```
klesgidis, line 41: Proposal: We could also add the producer name as a default header. WDYT?

src/initializers/kafka/kafka.ts
alkiskal: That's the partition key. I haven't exposed that argument

src/initializers/kafka/kafka.ts
nikosd23: +1 as this way we could filter on the client messages coming from different environments.

src/initializers/kafka/kafka.ts
nikosd23: maybe add is argument with default value null?

alkiskal: No need to expose it in this PR.

src/initializers/kafka/kafka.ts
alkiskal: I'm not sure about the filtering:
1. What about ruby apps?
2. What about node apps that are using older orka?

src/initializers/kafka/index.ts
```diff
@@ -20,3 +20,5 @@ export const getKafka = () => {
   }
   return kafka;
 };
+
+export { Kafka };
```
nikostoulas, line 24: This was used as typings. Exporting it might actually import the module in the produced .js file.  From what I see in L10 we didn't want that

src/initializers/kafka/kafka.ts
```diff
@@ -25,8 +25,21 @@ export default class Kafka {
     logger.info(`Kafka connected ${brokers.join(',')}`);
   }
 
-  public async send(topic: string, message: string | Buffer, key?: string, partition?: number) {
-    const { key: _key, partition: _partition, offset } = await this.producer.send(topic, message, partition, key);
+  public async send(
+    topic: string,
+    message: string | Buffer,
+    key?: string,
```
nikostoulas, line 31: We could add key, partition, headers as a third options argument.
This is backwards incompatible but it will be more readable in the future. 
We could add a couple of lines to make it backwards compatible if we need to eg.
```
if options is string 
   then key=options
```

alkiskal: Let's do this in another PR

src/initializers/kafka/kafka.ts
alkiskal: I will not be adding anything else in this PR. @klesgidis please open an enhancement request so we can discuss this.

Issue #11, opened 2019-05-21
Read visitor cookie id


nikostoulas:
This will be optional and used by apps that need it only

SokratisVidros:
That's right. I will fix it next week!

anastasios-workable:
FYI implemented @ https://github.com/Workable/application-form/blob/master/src/app/middlewares/visitor.ts

alkiskal:
what about this? should someone copy it in orka?

alkiskal:
I believe this can be closed @nikostoulas

nikostoulas:
Yeap. Closing.

PR #145, opened 2020-04-20
add visitor id as default middleware

alkiskal:
- Solves #11 

Notes:
- The original implementation includes logging and a call to newrelic
The log can be added, the newrelic call is somewhat troublesome, since we do not enforce a newrelic version it could cause problems. We can discuss if these two should be added or not.

### DIFF
--- a/src/initializers/diamorphosis.ts
+++ b/src/initializers/diamorphosis.ts
@@ -24,6 +24,7 @@ export default (config, orkaOptions: Partial<OrkaOptions>) => {
   config.port = config.port || 3000;
   config.allowedOrigins = config.allowedOrigins || ['localhost', 'lvh.me'];
   config.traceHeaderName = config.traceHeaderName || 'X-Request-Id';
+  config.visitor = config.visitor || { cookie: 'wmc' };
   config.blacklistedErrorCodes = config.blacklistedErrorCodes || [404];
   config.riviere = {
     enabled: true,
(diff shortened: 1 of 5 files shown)


src/initializers/diamorphosis.ts
```diff
@@ -24,6 +24,7 @@ export default (config, orkaOptions: Partial<OrkaOptions>) => {
   config.port = config.port || 3000;
   config.allowedOrigins = config.allowedOrigins || ['localhost', 'lvh.me'];
   config.traceHeaderName = config.traceHeaderName || 'X-Request-Id';
+  config.visitor = config.visitor || { cookie: 'wmc' };
```
nikostoulas, line 27: The name wmc isn't generic.

src/orka-builder.ts
```diff
@@ -65,6 +66,7 @@ export default class OrkaBuilder {
     );
     this.useCors();
     this.use(() => addRequestId(this.config));
+    this.use(() => addVisitorId(this.config));
```
nikostoulas, line 69: Is this is somehow optional? Not all apps need it. It should be optional.

src/initializers/koa/add-visitor-id.ts
```diff
@@ -0,0 +1,17 @@
...
+export default function(config): Middleware {
+  return async function addVisitorId(ctx: Context, next: () => void) {
+    switch (ctx.path) {
+      case '/health':
+        return await next();
+      default:
+    }
+    const cookie = ctx.cookies && ctx.cookies.get(config.visitor.cookie);
+    const { cookie_id: visitor } = decode(cookie);
+    ctx.state = { ...(ctx.state || {}), visitor };
```
nikostoulas, line 14:
```suggestion
    ctx.state.visitor = visitor;
```

PR #149, opened 2020-05-12
[CAN-2238] Add prefixes that return * for allowed origins

klesgidis:
## Summary
We need to serve some endpoints in careers-page with `access-control-allow-origin: *`

Endpoints accessible to everyone

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@ config = {
   cors: {
     allowedOrigins: ['localhost', 'lvh.me'],
     credentials: true, // Adds cors needed for exchanging cookies over https.
+    starPrefixes: ['/publicEndpoints/foo/bar'] // ctx.path prefixes that will return access-control-allow-origin: *
     …
   },
   traceHeaderName: 'X-Request-Id', // for logging in http requests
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -56,6 +56,7 @@ config = {
   cors: {
     allowedOrigins: ['localhost', 'lvh.me'],
     credentials: true, // Adds cors needed for exchanging cookies over https.
+    starPrefixes: ['/publicEndpoints/foo/bar'] // ctx.path prefixes that will return access-control-allow-origin: *
```
nikostoulas, line 59: Maybe allowed Prefixes?
```suggestion
    allowedPrefixes: ['/publicEndpoints/foo/bar'] // ctx.path prefixes that will return access-control-allow-origin: *
```

klesgidis: Not sure. Because what is `allowedPrefixes`? it is a bit misleading because we also have `allowedOrigins`.

Maybe `publicPrefixes`? WDYT?

nikostoulas: I like it better than starPrefixes

anastasios-workable: +1

PR #153, opened 2020-06-26
[CAN-2661] Separate kafka consumer/producer connection strings

klesgidis:
## Summary
Separates kafka consumer & producer configuration. From now on producer uses the following env variables:
* `KAFKA_PRODUCER_BROKERS`
* `KAFKA_PRODUCER_CERTIFICATES_KEY`
* `KAFKA_PRODUCER_CERTIFICATES_CERT`
* `KAFKA_PRODUCER_CERTIFICATES_CA`
* `KAFKA_PRODUCER_SASL_USERNAME`
* `KAFKA_PRODUCER_SASL_PASSWORD`

If `KAFKA_PRODUCER_BROKERS` is not provided then the global kafka options will be used for producer.  (`KAFKA_BROKERS`, `KAFKA_CERTIFICATES_KEY`, `KAFKA_CERTIFICATES_CERT`, `KAFKA_CERTIFICATES_CA`, `KAFKA_PRODUCER_SASL_USERNAME`, `KAFKA_SASL_PASSWORD`)

P.S. This change requires no amendments on the apps. They only need to bump orka and set the env variables (no need for config.js change).

### DIFF
--- a/src/initializers/diamorphosis.ts
+++ b/src/initializers/diamorphosis.ts
@@ -34,7 +34,7 @@ export default (config, orkaOptions: Partial<OrkaOptions>) => {
   };
   if (config.kafka) {
     config.kafka.producer = {
-      brokers: [],
+      brokers: [...config.kafka.producer?.brokers],
       certificates: {
         key: '',
         cert: '',
--- a/src/initializers/diamorphosis.ts
+++ b/src/initializers/diamorphosis.ts
@@ -2,6 +2,40 @@ import * as diamorphosis from 'diamorphosis';
 import { OrkaOptions } from '../typings/orka';
 import { defaultTo, isBoolean } from 'lodash';
 
+export const producerConfigOverrides = config => {
+  const {
+    KAFKA_PRODUCER_BROKERS,
+    KAFKA_PRODUCER_CERTIFICATES_KEY,
+    KAFKA_PRODUCER_CERTIFICATES_CERT,
+    KAFKA_PRODUCER_CERTIFICATES_CA,
+    KAFKA_PRODUCER_SASL_USERNAME,
+    KAFKA_PRODUCER_SASL_PASSWORD
+  } = process.env;
+
+  if (!KAFKA_PRODUCER_BROKERS) {
+    return {
+      ...config.kafka.producer,
+      brokers: config.kafka.brokers,
+      certificates: config.kafka.certificates,
+      sasl: config.kafka.sasl
+    };
+  }
+
+  return {
+    ...config.kafka.producer,
+    brokers: KAFKA_PRODUCER_BROKERS?.split(','),
+    certificates: {
+      key: KAFKA_PRODUCER_CERTIFICATES_KEY,
+      cert: KAFKA_PRODUCER_CERTIFICATES_CERT,
+      ca: KAFKA_PRODUCER_CERTIFICATES_CA
+    },
+    sasl: {
+      username: KAFKA_PRODUCER_SASL_USERNAME,
+      password: KAFKA_PRODUCER_SASL_PASSWORD
+    }
+  };
+};
+
 export default (config, orkaOptions: Partial<OrkaOptions>) => {
   config.nodeEnv = config.nodeEnv || 'development';
   config.app = {
@@ -34,6 +68,10 @@ export default (config, orkaOptions: Partial<OrkaOptions>) => {
   };
   diamorphosis(orkaOptions.diamorphosis);
   config.app.env = config.app.env || config.nodeEnv;
+  // Separate kafka producer/consumer connection strings
+  if (config.kafka) {
+    config.kafka.producer = producerConfigOverrides(config);
+  }
 
   if (config.log.console === '') {
     config.log.console = !config.log.json;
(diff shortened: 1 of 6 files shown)


src/initializers/diamorphosis.ts
```diff
@@ -2,6 +2,40 @@ import * as diamorphosis from 'diamorphosis';
...
+    return {
+      ...config.kafka.producer,
+      brokers: config.kafka.brokers,
+      certificates: config.kafka.certificates,
+      sasl: config.kafka.sasl
+    };
+  }
+
+  return {
+    ...config.kafka.producer,
```
nikostoulas, line 25: 🔧 What if someone has overwritten some values in his config.js?
You will overwrite them with the below code.

src/initializers/diamorphosis.ts
```diff
@@ -2,6 +2,40 @@ import * as diamorphosis from 'diamorphosis';
...
 
+export const producerConfigOverrides = config => {
+  const {
+    KAFKA_PRODUCER_BROKERS,
+    KAFKA_PRODUCER_CERTIFICATES_KEY,
+    KAFKA_PRODUCER_CERTIFICATES_CERT,
+    KAFKA_PRODUCER_CERTIFICATES_CA,
+    KAFKA_PRODUCER_SASL_USERNAME,
+    KAFKA_PRODUCER_SASL_PASSWORD
+  } = process.env;
```
nikostoulas, line 13: 🔧 Reading from process.env directly could cause some conflicts with diamorphosis. Eg someone could overwrite some values in config.js and be removed.

klesgidis: I chose not to go with config because in that case all the apps would need to change their config files as well. Not only bump orka. Do you want me to include something like:
```
certificates: {
      key: config.kafka.producer.certificates.key || KAFKA_PRODUCER_CERTIFICATES_KEY,
}
```?

Comment (nikostoulas) [reply src/initializers/diamorphosis.ts:13]: You could reuse the pattern we have with honeybadger :
```javascript
config.honeybadger = {
    apiKey: '',
    ...config.honeybadger
  };

if(config.kafka){
  config.kafka.producer ={
  brokers: [],
  certificates: {
      key:'',
      cert: '',
      ca: '',
     ...config.kafka?.producer?.certificates
    },
    sasl: {
      username: '',
      password: '',
     ...config.kafka?.producer?.sasl
    },
   ...config.kafka.producer
  }
}
```

before diamorphosis is called. And let diamorphosis do the replacements with the env.
After diamoprhosis is called you would then have to give defaults to these values eg:

```js
if(config.kafka){
  config.kafka.producer ={
  brokers: config.kafka.producer.brokers || config.kafka.brokers,
  certificates: {
      key: config.kafka.producer.certificates.key || config.kafka.certificates.key,
    },
   etc
  }
}
```
Possibly a defaultsDeep method from lodash could also help.
Wdyt?

klesgidis: I am ok with the first part, but  I can't do the 2nd part of your comment because imagine the following situation:
```
config: {
  kafka: {
    brokers: ['a', 'b']
    certificates: { key: '1', ca: '2', cert: '3'},
    producer: {
      brokers: ['c', 'd'],
      sasl: { username: 'asd', password: 'cvb'}
    }
  }
}
```
then the consumer will end up with the following config:
```
producer: {
      brokers: ['c', 'd'],
      certificates: { key: '1', ca: '2', cert: '3'},
      sasl: { username: 'asd', password: 'cvb'}
}
```
and it will use the `certificates` mechanism instead of `sasl`

klesgidis:
PTAL

src/initializers/diamorphosis.ts
```diff
@@ -32,9 +32,36 @@ export default (config, orkaOptions: Partial<OrkaOptions>) => {
...
+        ...config.kafka.producer?.sasl
+      },
+      ...config.kafka.producer
+    };
+  }
   diamorphosis(orkaOptions.diamorphosis);
   config.app.env = config.app.env || config.nodeEnv;
 
+  // Override kafka producer config with defaults if brokers is not set
+  if (config.kafka && (!config.kafka.producer?.brokers || config.kafka.producer.brokers.length === 0)) {
```
nikostoulas, line 56:
```suggestion
  if (config.kafka && config.kafka.producer?.brokers?.length === 0) {
```
or you could use isEmpty

klesgidis: this is not needed since it is checked here `!config.kafka.producer?.brokers`

klesgidis: If I change like you suggested if `brokers` is `undefined` it won't enter the `if` condition

nikostoulas: Yeap you are right. Ignore it.

test/diamorphosis/diamorphosis.test.ts
```diff
@@ -1,6 +1,7 @@
 import diamorphosis from '../../src/initializers/diamorphosis';
 import { OrkaOptions } from '../../src/typings/orka';
 import * as path from 'path';
+import { assert } from 'console';
```
nikostoulas, line 4: What is this?

PR #166, opened 2020-07-14
Allow orka users not to use visitor middleware on defaults

klesgidis:


src/orka-builder.ts
```diff
@@ -70,7 +70,9 @@ export default class OrkaBuilder {
     );
     this.useCors();
     this.use(() => addRequestId(this.config));
-    this.use(() => addVisitorId(this.config));
+    if (includeVisitor) {
```
anastasios-workable, line 73: @nikostoulas shall we remove it from the defaults altogether? Is anyone depending on it.

Otherwise, @klesgidis we should switch to use a config for this one, instead of an argument. How about `config.visitor.no_orka` or something like that.

nikostoulas: The middleware itself does the below check:   
 ```
if (ctx.path === '/health' || !config.visitor || !config.visitor.cookie) {
```

We could remove it from there and add it here. 
If config.visitor.cookie
If for some reason this doesn't work for you we could also use the config.visitor.enabled which is more explicit.

Comment (nikostoulas) [reply src/orka-builder.ts:73]: I think no-one else is using it. But making it easy to be used (eg with no code changes only an env config) would be nice.

Comment (anastasios-workable) [reply src/orka-builder.ts:73]: Let's go with either `config.visitor.enabled` or renaming the env variable `config.visitor.cookieName` ? I believe the middleware was forked from our codebase, so we ended up parsing the cookie twice, and the code itself is not inline now.

**Comment by klesgidis** [comment]
@anastasios-workable @nikostoulas after this change the middleware will not be used by default. Since this is not used by others, it is not an issue

File: src/orka-builder.ts [hunk 2]
```diff
@@ -70,7 +70,9 @@ export default class OrkaBuilder {
     );
     this.useCors();
     this.use(() => addRequestId(this.config));
-    this.use(() => addVisitorId(this.config));
+    if (this.config.visitor && this.config.visitor.orkaEnabled) {
```
nikostoulas, line 73: 🙃 I prefer enabled instead of orkaEnabled

klesgidis: I though of writing this as `enabled` but I think that will be misleading for the apps that use orka. Do you want to call it just `orka`?

anastasios-workable: ditto

PR #172, opened 2020-07-17
WIP Add bull to orka

MariaEmma:
This is the first approach to add bull to orka. 
Since some apps need redis only for bull, we can declare the redis url and options in the config.bull, if the app uses redis we can just have it at first level.
One thing we should decide is how we will declare the queues and their options, in the approach here, we declare it in the config file in queues array.

### DIFF
--- a/src/initializers/bull/base-queue.ts
+++ b/src/initializers/bull/base-queue.ts
@@ -1,0 +1,67 @@
+import Queue = require('bull');
+import { getLogger } from 'log4js';
+import registry from './queue-registry';
+
+export default class BaseQueue {
+  queue: Queue.Queue<any>;
+  _logger: any;
+  static _initialize(name, options): Queue.Queue {
+    const queue = registry.getQueue(name, options);
+    return queue;
+  }
+
+  constructor(options = {}) {
+    this.queue = (this.constructor as typeof BaseQueue)._initialize(
+      this.name(),
+      options
+    );
+  }
+
+  name(): string {
+    return 'base';
+  }
+
+  get(): Queue.Queue {
+    return this.queue;
+  }
+
+  get logger(): any {
+    if (!this._logger) {
+      this._logger = getLogger(`queue.${this.name()}`);
+    }
+    return this._logger;
+  }
+
+  // Override these methods in subclass
+  async processor(data): Promise<any> {}
+  async beforeProcess(): Promise<any> {}
+
+  process(): Promise<any> {
+    return Promise.resolve()
+      .then(() => this.beforeProcess())
+      .then(() =>
+        this.queue.process((job) => {
+          this.logger.info(
+            `[${this.queue.name}] Processing job ${job.id}: ${JSON.stringify(
+              job.data
+            )}`
+          );
+          return this.processor(job.data);
+        })
+      );
+  }
+
+  add(job: any, options = {}): Promise<any> {
+    return this.queue.add(job, options);
+  }
+
+  static add(job: any, options = {}): Promise<any> {
+    const queue = new this();
+    return queue.add(job, options);
+  }
+
+  static process(): Promise<any> {
+    const queue = new this();
+    return queue.process();
+  }
+}
--- a/src/initializers/bull/index.ts
+++ b/src/initializers/bull/index.ts
@@ -1,0 +1,53 @@
+import { list, decode } from './helper';
+import { start } from './bull';
+import { getStats, reportStats } from './metrics';
+import { parseURL } from 'ioredis/built/utils';
+import defaults = require('lodash/defaults');
+import { getLogger } from '../log4js';
+import * as cron from 'node-cron';
+
+export default async (config) => {
+  const redisUrl = config.bull.redis.url || config.redis.url;
+
+  if (!config.bull && !redisUrl) {
+    return;
+  }
+  const defaultQueueOptions = {
+    redis: {
+      ...parseURL(redisUrl),
+      tls: decode(config.bull.redis.options.tls),
+    },
+    defaultJobOptions: {
+      removeOnComplete: true,
+    },
+  };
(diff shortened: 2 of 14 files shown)


src/initializers/bull/bull-helper.ts
```diff
@@ -0,0 +1,22 @@
+import * as path from 'path';
```
nikostoulas, line 1: 🙃  rename the fie to helper.ts

src/initializers/bull/bull-metrics.ts
```diff
@@ -0,0 +1,42 @@
+import { list, queuePath } from './bull-helper';
```
nikostoulas, line 1: 🙃  rename the file to metrics.ts

src/initializers/bull/bull-metrics.ts
```diff
@@ -0,0 +1,42 @@
...
+import { getLogger } from '../log4js';
+import * as newrelic from 'newrelic';
```
nikostoulas, line 3: 🔧  This might be required before the application has started. Some application might not have newrelic installed it's an optional peer-depencency. 
Use this instead https://github.com/Workable/orka/blob/338db674819e9bdc5bedf227a50669509e8fb9c6/src/initializers/newrelic/index.ts#L13

src/initializers/bull/bull-metrics.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+  return stats;
+}
+
+export async function reportStats(config): Promise<void> {
+  const stats = await getStats(config);
+
+  // tslint:disable-next-line: forin
+  for (const queueName in stats) {
+    newrelic.recordMetric(
```
nikostoulas, line 29: 🔧  Check if new relic is available first. See comment above

src/initializers/bull/base-queue.ts
```diff
@@ -0,0 +1,68 @@
+require('../../config/initializer');
```
nikostoulas, line 1: ❓ what is this?

src/initializers/bull/bull-helper.ts
```diff
@@ -0,0 +1,22 @@
...
+  let queueNames = [];
+  if (config.bull.queues) {
+    queueNames = config.bull.queues.map((q) => q.name);
+  }
+  return queueNames;
+}
+
+export function encode(val): any {
+  isNil(val) ? val : Buffer.from(JSON.stringify(val)).toString('base64');
+}
```
anastasios-workable, line 18: Nil and undefined values are falsy in javascript, so I don't think you need the `lodash.isNil`.

You could use javascripts's native `atob()` and `btoa()` methods. Also did you forget to return in this method? And keep in mind that that stringifying a string adds extra quotes, is that what you want?

May I suggest

```suggestion
export const encode = (val: any) =>  val && btoa(`${val}`);
```

src/initializers/bull/index.ts
```diff
@@ -0,0 +1,51 @@
+import { list, decode } from './bull-helper';
+import { start } from './bull';
+import { getStats, reportStats } from './bull-metrics';
+import { parseURL } from 'ioredis/built/utils';
+import defaults = require('lodash/defaults');
+import { getLogger } from '../log4js';
+import * as cron from 'node-cron';
```
anastasios-workable, line 7: Shall we make this dependency also optional? @nikostoulas

src/initializers/bull/index.ts
```diff
@@ -0,0 +1,51 @@
...
+  const queues = await list(config);
+  for (const queueName of queues) {
+    const queueOptions = config.bull.queues.find((q) => q.name === queueName)
+      .options;
+    const opts = defaults({}, queueOptions, defaultQueueOptions);
+    await start(queueName, opts);
+  }
+};
+
+export const printStats = async (config): Promise<void> => {
```
anastasios-workable, line 37: return type is redundant since typescript infers it from the code.

```suggestion
export const printStats = async config => {
```

src/initializers/bull/index.ts
```diff
@@ -0,0 +1,51 @@
...
+};
+
+export const reportMetrics = (config): any => {
+  if (config.newRelicStatsCronExpression) {
+    cron.schedule(config.newRelicStatsCronExpression, async () => {
+      getLogger('queue.cli').info('Sending queue stats in new relic');
+      await reportStats(config);
+    });
+  } else {
+    return 0;
```
anastasios-workable, line 49: ?

src/initializers/bull/index.ts
nikostoulas: This and bull could be made optional. This seems light with not many dependencies. So it might not worth it. Bull on the other hand might worth it. I don't have any strong opinion. 
@MariaEmma note that if they are not optional their vulnerabilities will affect all our projects.

src/initializers/bull/bull-metrics.ts
MariaEmma: Fixed it.Thank you!

src/initializers/bull/base-queue.ts
MariaEmma: Oupsss..removed it.

src/initializers/bull/index.ts
MariaEmma: You are right, we should make it optional, I will do the changes.

src/initializers/bull/index.ts
MariaEmma: I log a message to make it more clear!thanks for noticing.

src/initializers/bull/bull-helper.ts
MariaEmma: Thank you for the solution, I removed the method completely since I don't use it after all.

echetzakis:
Closing this PR in favour of #177

PR #177, opened 2020-08-12
Adds Bull to Orka

echetzakis:
### Description
This feature brings [Bull Queues](https://github.com/OptimalBits/bull) in the Orka framework.
Heavily inspired (and partly copy-pasted) from #172.

### Why a new PR? 
This PR takes a step back (compared with #172) providing a more low level abstraction over bull, rather that the more opinionated approach we've implemented in `VHS`.

This way it seems much easier to use in other projects (including VHS) without necessarily following the same approach as we have in VHS, and with smaller code changes.

### Usage
For details see the relevant section added to `README.md` and the `bull-example` provided.

### Design doc
https://docs.google.com/document/d/1aRPCmmLFrggjfRAGrer_LaoP2DoABq4BvxW7lZ35t7U


src/initializers/bull/bull.ts
```diff
@@ -0,0 +1,121 @@
...
+  private instances = {};
+  private metricsTask: any;
+
+  constructor(prefix, queueOpts, defaultOptions, redisOpts) {
+    this.prefix = prefix;
+    this.defaultOptions = defaultOptions;
+    this.queueOpts = queueOpts.reduce((acc, q) => {
+      acc[q.name] = q;
+      return acc;
+    }, {});
```
anastasios-workable, line 23: nit since you use lodash:

```suggestion
    this.queueOpts = keyBy(queueOpts, 'name');
```

src/initializers/bull/bull.ts
```diff
@@ -0,0 +1,121 @@
...
+    const q = this.getQueue(name);
+    return {
+      queue: name,
+      count: await q.count(),
+      failed: await q.getFailedCount()
+    };
+  }
+
+  public getStats() {
+    const stats = Object.keys(this.queueOpts).map(name => this.fetchStats(name));
```
anastasios-workable, line 67: nit: 
```suggestion
    const stats = Object.keys(this.queueOpts).map(this.fetchStats);
```

src/initializers/bull/index.ts
```diff
@@ -0,0 +1,44 @@
...
+    if (isEmpty(config.bull.redis.tls)) delete config.bull.redis.tls;
+  }
+  const redisOpts = {
+    ...parseURL(redis.url),
+    tls: redis.tls || redis.options?.tls,
+    enableReadyCheck: false
+  };
+  const Bull = (await import('./bull')).default;
+  bull = new Bull(prefix, queues, defaultOptions, redisOpts);
+  getLogger('orka.initializers.bull').info(
```
anastasios-workable, line 31: Here as well, move the getLogger statements outside the function scope. Supposedly loggers are cached, but there is no need to invoke this constantly.

test/initializers/bull/bull.test.ts
```diff
@@ -0,0 +1,125 @@
...
+  const defaultOptions = { removeOnComplete: true };
+  const redisOptions = { url: 'redis://localhost:6379/' };
+  let bull;
+
+  afterEach(() => {
+    sandbox.restore();
+  });
+
+  beforeEach(async () => {
+    // scheduleMock = sandbox.stub();
```
anastasios-workable, line 20: remove

src/initializers/bull/bull.ts
echetzakis: your suggestion does not work unless you do a `this.fetchStats.bind(this)` which in turn raises a TS error because the return type of the bound function cannot be inferred anymore

anastasios-workable: Try:

```
const stats = Object.keys(this.queueOpts).map(this.fetchStats, this);
```

src/initializers/bull/index.ts
echetzakis: but of course! map accepts a `thisArg`!
ty! 🙇

PR #182, opened 2020-09-08
Improvements for Bull integration

echetzakis:
This PR:
* Allows passing more options to Bull's redis connection (ie `enableReadyCheck`)
* Allows to customise the queue name case in custom metrics reported to newrelic (in order not to break existing dashboards/alerts if the naming was not camelCase)

### DIFF
--- a/src/initializers/bull/index.ts
+++ b/src/initializers/bull/index.ts
@@ -1,4 +1,4 @@
-import { isEmpty } from 'lodash';
+import { isEmpty, omit } from 'lodash';
 import { getLogger } from '../log4js';
 import { OrkaOptions } from '../../typings/orka';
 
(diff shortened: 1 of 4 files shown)


src/initializers/bull/index.ts
```diff
@@ -14,6 +14,9 @@ export default async (config, orkaOptions: Partial<OrkaOptions>) => {
   const prefix = orkaOptions.appName;
   const defaultOptions = config.bull.queue.options;
   const redis = config.bull.redis || config.redis;
+  const options = defaults(omit(config.bull.redis || config.redis.options, ['url', 'tls']), {
```
anastasios-workable, line 17: No need for defaults here, the spread operator would suffice, no?

```suggestion
  const options = {...omit(config.bull.redis || config.redis.options, ['url', 'tls']), enableReadyCheck: false};
```

anastasios-workable (approved):

src/initializers/bull/index.ts
echetzakis: Your suggestion does not equal the `defaults` behaviour because it will always set `enableReadyCheck: false`.

The intention here is to set as default the `enableReadyCheck: false`, but to be overridden if needed via the configuration

anastasios-workable: So?:

```
const options = {enableReadyCheck: false., ..omit(config.bull.redis || config.redis.options, ['url', 'tls'])};
```

PR #196, opened 2020-11-19
[ENG-18] Add request context middleware to orka

klesgidis:
## Summary
This is the first PR of adding a `Request Context` to orka.

Example for how async_hooks work: https://www.icloud.com/keynote/0uwcgADQfQRGhUDkzAZ4SMd7A#AsyncLocalStorage

This change requires NodeJS: 12.17.0+, 13.14.0+, or 14.0.0+

Another PR will follow:
* Reading this context and exposing those variables to the log4js appenders
* Adding an exampe in the `/examples` folder

### DIFF
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -4,6 +4,10 @@ import _defaults from './default-options';
 import * as lodash from 'lodash';
 import { getLogger } from './initializers/log4js';
 import orkaType from './orka-builder';
+import { AsyncLocalStorage } from 'async_hooks';
+
+const als = new AsyncLocalStorage<Map<string, any>>();
+export const getRequestContext = () => als.getStore();
 
 export default (defaults: Partial<OrkaOptions> = _defaults) => {
   const options: Partial<OrkaOptions> = lodash.cloneDeep(lodash.defaultsDeep({}, defaults, _defaults));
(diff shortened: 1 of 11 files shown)


nikosd23 (commented):
🥇

src/builder.ts
```diff
@@ -4,6 +4,10 @@ import _defaults from './default-options';
 import * as lodash from 'lodash';
 import { getLogger } from './initializers/log4js';
 import orkaType from './orka-builder';
+import { AsyncLocalStorage } from 'async_hooks';
```
nikostoulas, line 7: ❓ Will this break if used with node < 12.17 ?

src/initializers/koa/add-visitor-id.ts
```diff
@@ -14,8 +15,9 @@ export default function(config): Middleware {
     try {
       const { cookie_id: visitor } = decode(cookie);
       ctx.state.visitor = visitor;
+      getRequestContext()?.set('visitor', ctx.state.visitor);
```
nikostoulas, line 18: 👏

nikostoulas (approved):
Nicely done @klesgidis! I was wondering if this is a breaking change (and should bump our major version) or if there is a way to make it non breaking.

src/builder.ts
klesgidis: Yes this will break if node < 12.17

We could go for `v1.0.0`

klesgidis: I could also enhance readme with the minimum NodeJS version

echetzakis (approved):
👌  nice!

PR #198, opened 2020-11-24
[ENG-18] Do not crash if appropriate NodeJS version is not installed

klesgidis:
## Summary
If appropriate nodeJS version is installed, include `requestContext` by default, otherwise do not include it and show a warning.

P.S. If user overrides configuration by hand the app will crash


src/utis.ts
```diff
@@ -0,0 +1,27 @@
+const regex = /^v(\d+)\.(\d+)\.(\d+)$/;
```
nikostoulas, line 1: 👍  +1 for doing this here. I wouldn't like adding a dependency for this only.

src/orka-builder.ts
```diff
@@ -67,6 +68,10 @@ export default class OrkaBuilder {
     this.use(() => addRequestId(this.config));
     if (this.config.requestContext.enabled) {
       this.use(() => addRequestContext(this.als));
+    } else if (!alsSupported()) {
+      getLogger('orka').warn(
+        'RequestContext is disabled!! You need to have version v12.17.0+, v13.14.0+ or v14.0.0+ if you want to enable it.'
```
nikostoulas, line 73: 👏

nikostoulas (approved):
💯 thank you for doing this @klesgidis . Excellent work.

src/utis.ts
nikosd23, line 1: nit: There is a missing L from the name of the file.

nikosd23 (approved):

src/utis.ts
klesgidis: nice catch

PR #200, opened 2020-11-26
Adds datadog in orka

nikosd23:


src/initializers/datadog/index.ts
```diff
@@ -0,0 +1,19 @@
+import requireInjected from '../../require-injected';
+import { getLogger } from '../log4js';
+
+let tracer;
+export default config => {
+  if (process.env.DD_SERVICE && process.env.DD_ENV) {
+    tracer = requireInjected('dd-trace').init();
+    tracer.use('koa', {
```
stavros-wb, line 8: maybe we can have an optional config with an array of datadog plugins to be enabled?

stavros-wb (approved):

klesgidis (approved):
Nice!!
Maybe add the config option that enables datatog in readme (like newrelic)

src/initializers/datadog/index.ts
```diff
@@ -0,0 +1,19 @@
...
+      blacklist: config?.datadog?.blacklistedPaths?.split(',') || ['/health']
```
nikostoulas, line 9: 🔧 Initialise datadog.blacklistedPaths = ['/health'] in diamorphosis and then diamorphosis does the split.
```suggestion
      blacklist: config?.datadog?.blacklistedPaths
```
Maybe name it `dd.blacklistedPaths` to match DD_SERVICE and DD_ENV?
And in env: DD_BLACKLISTED_PATHS if it needs to change through an env?

jgerbe (approved):

nikostoulas (approved):

src/initializers/datadog/index.ts
nikosd23: I don't want to use DD_* as this convention is used by the dd-tracer itself.

src/initializers/datadog/index.ts
nikosd23: According to the API found [here](https://github.com/DataDog/dd-trace-js/blob/master/docs/API.md)  the instrumentation of external libraries using the built-in plugins is done automatically. You can also configure disabled plugins through env variables which is the recommended way to configure datadog, that's why I kept the initialisation simple.

kzacharakis (approved):

PR #213, opened 2020-12-16
Add github scripts for testing, releasing and publishing to npm

nikostoulas:
Move unit testing to github (our plan now supports actions for open source projects)

Simplify creating a release. For those that have [admin](https://github.com/orgs/Workable/teams/node-js-admins) access the process will be as follows:
- `git checkout master && git pull`
- `npm version minor`
- `git push && git push --tags`
- wait for tests to pass and release to be automatically created
- Go to the release page and write a proper release description
- Wait for the job to publish orka to npm

For non-admins the tag creations should continue to happen through a pr that then is *merged* (not squashed) to master.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,4 +1,4 @@
-const regex = /^v(\d+)\.(\d+)\.(\d+)$/g;
+const regex = /^v(\d+)\.(\d+)\.(\d+)$/;
 
 export const alsSupported = () =>
   nodeVersionGreaterThanEqual('v12.17.0') ||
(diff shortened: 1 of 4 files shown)


src/utils.ts
```diff
@@ -1,13 +1,17 @@
...
   nodeVersionGreaterThanEqual('v13.14.0') ||
   nodeVersionGreaterThanEqual('v14.0.0');
 
 export const nodeVersionGreaterThanEqual = (requestedVersion: string, version = process.version) => {
-  const [major, minor, patch] = [...version.matchAll(regex)][0].slice(1).map((n: string) => parseInt(n, 10));
-  const [requestedMajor, requestedMinor, requestedPatch] = [...requestedVersion.matchAll(regex)][0]
+  const [major, minor, patch] = version
+    .match(regex)
+    .slice(1)
+    .map((n: string) => parseInt(n, 10));
```
klesgidis, line 12: 👏 Nice! 
nit: 
```suggestion
  const [, major, minor, patch] = version
    .match(regex)
    .map((n: string) => parseInt(n, 10));
```

klesgidis (approved):

src/utils.ts
nikostoulas: Done. @klesgidis we have one more change that needs to be done in order to support Node10 too. We need an alternative to Object.fromEntries. See https://github.com/feross/fromentries/blob/master/index.js

PR #214, opened 2020-12-17
[ENG-41] Add default option to riviere to blacklist DD traces

klesgidis:
Related: https://github.com/Workable/riviere/pull/71

### DIFF
--- a/src/initializers/koa/riviere.ts
+++ b/src/initializers/koa/riviere.ts
@@ -16,8 +16,7 @@ export default (config, orkaOptions: Partial<OrkaOptions>) =>
       https: true,
       level: 'info',
       maxBodyValueChars: config.riviere.maxBodyValueChars,
-      blacklistedPaths: ['/v0.4/traces'],
-      ...config.riviere.outbound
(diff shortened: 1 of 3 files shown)


src/initializers/koa/riviere.ts
```diff
@@ -15,15 +15,18 @@ export default (config, orkaOptions: Partial<OrkaOptions>) =>
       enabled: config.riviere.enabled,
       https: true,
       level: 'info',
-      maxBodyValueChars: config.riviere.maxBodyValueChars
+      maxBodyValueChars: config.riviere.maxBodyValueChars,
+      blacklistedPaths: ['/v0.4/traces'],
```
nikosd23, line 19: A regex support would be more future proof but let's iterate on a next version of reviere

nikosd23 (approved):
💯

src/initializers/koa/riviere.ts
klesgidis: That's a good idea. I will convert it now, on both PRs

PR #220, opened 2021-01-28
Kafka upgrade to latest Kafka.js

nikostoulas:
I decided to drop sinek as it is in alpha version for the last 3 months and it uses kafkajs under the hood.
The documentation of kafka.js is extremely useful making it a good choice see [here](https://kafka.js.org/docs/introduction)
This pr:

- Drops sinek
- Adds kafka.js in an almost fully compatible way, but:
  - Kafka.send is deprecated (but still supported) in favour of Kafka.publisher.send which is the [native kafka.js send](https://kafka.js.org/docs/producing)
  - Kafka.metadata() is turned into an async method and its output possibly changes-> I don't believe it was documented anywhere though, possibly only used in bekafka
  - Two type changes when consuming messages:
    - timestamp is now a string (not sure why kafka.js changed this)
    - offset is also a string due to the fact that offset's can be larger than what a JS number can represent.
- Added support for an easy local kafka broker with the below command: 
```
npm explore @workable/orka -- npm run kafka:start
````
- added support for creating topics programmaticaly:
```
getKafka().createTopics([
  { topic: 'foo', numPartitions: 10, replicationFactor: 1 },
  { topic: 'bar', numPartitions: 10, replicationFactor: 1 },
  { topic: 'test', numPartitions: 10, replicationFactor: 1 }
]);
```
It will not complain if a topic already exists. Will create any topics added and log what it created vs was already there.

Will run the suite in our apps to make sure that nothing unexpected breaks before merging. 

Leaning towards a minor version upgrade of orka. Wdyt?

Haven't updated README. I am addressing readme and orka docs in a different pr coming soon.

> :warning:  In officedroid it required a hard restart. Old pods to be killed before new ones were up because it seems that kafka.js is using a different partitionAssigners algorithm (by default [PartitionAssigners.roundRobin](https://kafka.js.org/docs/consuming#a-name-custom-partition-assigner-a-custom-partition-assigner)) for consumers.

### CODE REVIEW

File: src/initializers/kafka/base-kafka-handler.ts [hunk 1]
```diff
@@ -1,52 +1,93 @@
...
+            this.logger.error(e);
+          }
+        }
+        if (this.stringifyHeaders && message.headers) {
+          message.rawHeaders = message.headers;
+          message.headers = Object.keys(message.headers).reduce(
+            (m, k) => ({ ...m, [k]: message.headers[k].toString() }),
+            {}
+          );
+        }
```
nikostoulas, line 88: Needed for backwards compatibility. Value, headers are buffers in kafka.js

package.json
```diff
@@ -82,8 +84,9 @@
     "@types/redis": "^2.8.14",
     "@types/sinon": "^7.5.0",
     "bull": "*",
-    "dd-trace": "*",
+    "dd-trace": "^0.30.5",
```
nikosd23, line 87: did you intend to change this with this pr?

nikosd23 (approved):
🎉

package.json
nikostoulas: No, nice catch!
I am reverting it even though it is devDependencies (It won't affect apps using orka)

kzacharakis (approved):
🎖️

echetzakis (approved):

jgerbe (approved):
:

src/initializers/kafka/kafka.ts
```diff
@@ -1,100 +1,134 @@
...
+          .then(response => ({ [topic.topic]: response }))
+      )
+    );
+    await admin.disconnect();
+    logger.info(`Created topics: ${JSON.stringify(responses)}`);
+    return responses;
+  }
+
+  /**
+   * @deprecated  use producer.send instead which is native kafkajs send
```
klesgidis, line 101: 👏

src/initializers/kafka/kafka.ts
```diff
@@ -1,100 +1,134 @@
...
     topic: string,
     message: string | Buffer,
-    key?: string,
+    key: string = uuid.v4(),
     partition?: number,
-    headers?: { [key: string]: string }[]
+    inputHeaders?: { [key: string]: string }[]
   ) {
-    const { key: _key, partition: _partition, offset } = await this.producer.send(
+    if (!key) key = uuid.v4();
```
klesgidis, line 110: This is redundant

src/initializers/diamorphosis.ts
```diff
@@ -75,23 +75,9 @@ export default (config, orkaOptions: Partial<OrkaOptions>) => {
...
-      },
-      sasl: {
-        username: '',
-        password: '',
-        ...config.kafka.producer?.sasl
-      },
-      ...config.kafka.producer
-    };
-  }
+
```
klesgidis, line 78: I think that we had  the `if (config.kafka) {` to not initialize kafka if it was missing from config. Is this change going to initialize kafka even if it is not present in the config file?

klesgidis (approved):
Nice work!

src/initializers/kafka/kafka.ts
nikostoulas: does kafka.js add a key by default? Because our previous implementation did so.

nikostoulas: I added it because we had tests that set key as null.

src/initializers/diamorphosis.ts
nikostoulas: Since I initialise it as an empty string and brokers as an empty array those will still be evaluated to false.
My logic here is that on application level you won't  even need to set those variables on config. You could directly add them as env variables (in .env in development) and still work. Less configuration is better.

My idea is to do the same with the rest too, mongodb, rabbitMQ and redis. WDYT?

klesgidis: Agree

src/initializers/kafka/kafka.ts
klesgidis: No it doesn't add it by default but it will have a value from this:
`key: string = uuid.v4(),` (line 106)

nikostoulas: Unless someone sets key:null as we did in the tests in order to also add headers for example.

nikostoulas:
Good news: Our suite has passed, so I am merging this.
Bad news: We can no longer have multiple consumers with the same groupId but consuming from different topics. Each consumer group has to consume from exactly the same topics, so we will possibly have to rename some consumer groups in production. A new pr will follow to simplify this task.
(CC @innnkkki  we will be renaming kafka group ids in production in the near future)

PR #221, opened 2021-02-01
Rename group ids

nikostoulas:
**TLDR**: Kafka.js does not allow multiple consumers with the same groupId. Apps that have > 1 topics currently with node-rdkafka will have to rename all (but one) groupIds. 

This pr introduces a convenient way to do this in production:

```js
await getKafka().renameGroupId([
  { groupId: config.kafka.groupId, topic, oldGroupId: config.kafka.oldGroupId },
  { groupId: config.kafka.groupId2, topic, oldGroupId: config.kafka.oldGroupId2 },
  ...rest
]);
```

 In kafka.js: consumers with the same groupId must subscribe to exactly the same topics. See [RECEIVED_UNSUBSCRIBED_TOPICS](https://kafka.js.org/docs/faq#is-kafkajs-used-in-production)
I have also asked kafkajs [slack channel](https://kafkajs.slack.com/archives/CF6RFPF6K/p1611943584031700) and the suggested way forward is to rename our groupIds to sth unique.

The way we previously used orka all consumers of an app had the same groupId with no way to configure it. 
There is an easy way now to configure different groupIds per consumer:

```js
new KafkaHandler(
  kafka,
  { topic, consumerOptions: { groupId: config.kafka.groupId + '.' + options.topic } }
);
```

### DIFF
--- a/examples/kafka-example/app.js
+++ b/examples/kafka-example/app.js
@@ -1,4 +1,4 @@
-const { orka } = require('../../build');
+const { orka, getKafka } = require('../../build');
 
 orka({
   beforeMiddleware: () => [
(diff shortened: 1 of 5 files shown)


nikosd23 (approved):

kzacharakis (approved):

klesgidis (approved):
Should we have a migration process for all the apps? What we'll need to do to migrate?

jgerbe (approved):
groupid rename shall affect dd alerts as well

nikostoulas:
> Should we have a migration process for all the apps? What we'll need to do to migrate?

I will add a section in documentation, but it's as simple as in the description of the pr.

> groupid rename shall affect dd alerts as well

Indeed.

PR #222, opened 2021-02-03
Introduce documentation

nikostoulas:
📜  I have refactored readme and moved most documentation under /docs folder. 📜 
I have added more docs myself, including for kafkajs that was just developed.

🔧  While adding some docs, I saw that I had to align some configuration initialisation to match the general direction of orka. (Convention over configuration)

1️⃣  This is a first attempt. Having a dedicated page was pretty easy and well supported by github.
We have also talked with marketing to give us a better logo and/or styling.

📄 Docs need a lot more love and contribution from everyone. Reading and correcting mistakes will help a lot.
Whenever someone can't figure out how sth is used or works we should update the docs too to include the knowledge.

👁️  You can check the results [here](https://workable.github.io/orka). They are now served from 'gh-pages' branch but once this is merged I will change it to be served from master.


nikosd23 (approved):
🥇

klesgidis (approved):
💯 This is awesome

PR #227, opened 2021-02-23
Fix application name configuration SPC-244

manoskouvarakis:


nikostoulas (approved):
👏

test/builder.test.ts
```diff
@@ -12,4 +16,11 @@ describe('builder', function() {
   it('appends middleware koa');
   it('prepends middleware koa');
   it('uses port');
+
+  it('should get app name from config', () => {
```
nikostoulas, line 20: 🙃  can you add a test where options.appName is set?

manoskouvarakis: Done

PR #231, opened 2021-03-23
[ENG-74] [PROD-1676] Add runWithContext function

klesgidis:
## Summary
* Exports typings to orka's getLogger
* Creates a new function called `runWithContext` to be used by workers, handlers etc. This function can be used to execute a callback inside it that will have an available request context. (see usage in the documentation)

### DIFF
--- a/docs/request-context.md
+++ b/docs/request-context.md
@@ -51,7 +51,7 @@ e.g.
 import { getRequestContext } from '@workablehr/orka';
 ```
 
-### Examples
+#### Usage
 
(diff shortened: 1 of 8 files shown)


nikostoulas (approved):
❓ Will orka support by default rabbitmq and kafka?

klesgidis:
> ❓ Will orka support by default rabbitmq and kafka?

Discussed offline and I will take advantage of this functionality on Kafka & Rabbit handlers in a future PR

PR #233, opened 2021-03-25
Update mongoose to fix vulnerability of mquery

klesgidis:


jgerbe (approved):

nikostoulas (approved):
❓ Why are tests failing @klesgidis

klesgidis:
> ❓ Why are tests failing @klesgidis

that's a valid question 😛 it is something with the new types

PR #235, opened 2021-03-29
[PROD-1677][ENG-75] Automatically pass trace information on request headers

klesgidis:
## Summary
* Always initialize riviere on orka builder
* Add new HTTP proxy that automatically appends trace id to all requests, if exists


nikostoulas (approved):
💯

nikostoulas:
I was wondering if we should rename the default traceHeaderName(maybe x-orka-id) as istio overwrites x-request-id. WDYT?

klesgidis:
> I was wondering if we should rename the default traceHeaderName(maybe x-orka-id) as istio overwrites x-request-id. WDYT?

this will affect all apps, and won't be backwards compatible. I don't have a problem renaming it, but maybe we should contact code owners?

nikostoulas:
> > I was wondering if we should rename the default traceHeaderName(maybe x-orka-id) as istio overwrites x-request-id. WDYT?
> 
> this will affect all apps, and won't be backwards compatible. I don't have a problem renaming it, but maybe we should contact code owners?

We haven't released v2 yet so that's ok. In any case this changes through env configuration. This hasn't to do with this pr, so feel free to merge this.

nikosd23 (commented):
🥇

PR #247, opened 2021-04-23
On consumer created callback

klesgidis:
Adds callback for kafka to be used by clients when kafka consumer is created


nikostoulas (approved):
❓ What's the use case for this @klesgidis ?

nikostoulas:
Can we also update documentation for this?

klesgidis:
> ❓ What's the use case for this @klesgidis ?

Bekafka of course 😛 [Code](https://github.com/Workable/bekafka-api/blob/master/server/src/app/consumers/fetch-consumer.ts#L27-L37)

I'll update the docs

PR #248, opened 2021-04-26
Do not try connect to kafka even on empty strings array

manoskouvarakis:


klesgidis (approved):

nikostoulas (approved):
Nice @manoskouvarakis .
Maybe we should also update https://github.com/Workable/diamorphosis to handle empty strings as empty arrays not as arrays with an empty string.

PR #257, opened 2021-06-03
Provide an easy way through config to controll wether to use ssl or not: (useSsl) by default true.

nikostoulas:
Currently we cannot add in config.postgres.sslConfig a default value that can be overwritten by each env. If provided postgres client will always try to use ssl.
So config.postgres.useSsl is added with this pr. This will default to true (that works well with all with all envs except local).


klesgidis (approved):
👍

PR #263, opened 2021-09-08
[ENG-95] Log kafka expected errors as warnings

klesgidis:
## Summary
This will log errors like [this](https://app.honeybadger.io/projects/60297/faults/78990832) as warning.

This way we will get alerted only for non expected errors in the consumer.


nikostoulas (approved):
👏

jgerbe (approved):
🔇

PR #264, opened 2021-09-10
Introduce a worker factory

nikostoulas:
We use a similar worker in one of our services.
It uses a mongo collection to save the state of the worker so that when it starts again it will continue from where it had left off.
This seems to be useful in other services too so I am adding it here.
It is mostly boilerplate code that initializes all orka tasks for the worker to have them available.
Also it does the most of the logging - mongodb status updates.

### DIFF
--- a/docs/extra/workers.md
+++ b/docs/extra/workers.md
@@ -15,7 +15,7 @@ nav_order: 1
   {:toc}
 
 Orka is exposing a useful factory to create workers. This is a convenient boilerplate code to execute long running tasks keeping some state in a mongodb collection.
-Sometimes you need to execute long running tasks that need to continue from where the left when being restarted. E.g. Processing some million of rows in a table.
+Sometimes you need to execute long running tasks that need to continue from where it had left off when being restarted. E.g. Processing some million of rows in a table and keeping the last processed row.
 Orka's worker will automatically create and use a mongo collection called workerjob for this.
 
 ## Installation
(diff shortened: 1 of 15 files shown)


docs/extra/workers.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Table of contents
+{: .no_toc .text-delta }
+
+1. TOC
+  {:toc}
+
+Orka is exposing a useful factory to create workers. This is a convenient boilerplate code to execute long running tasks keeping some state in a mongodb collection.
+Sometimes you need to execute long running tasks that need to continue from where it had left off when being restarted. E.g. Processing some million of rows in a table and keeping the last processed row.
+Orka's worker will automatically create and use a mongo collection called workerjob for this.
```
nikosd23, line 19: it will be useful to add an example of how this integrates with a cloud environmnet. ie, how we used it in k8s or how one can use let;s say using another cloud provider.

nikosd23 (approved):
🥇

manoskouvarakis (approved):

docs/extra/workers.md
nikostoulas: Is our solution with production-env-configs something that makes sense outside workable? 
Also we currently don't have any cloud environment documentation for starting an orka enabled app.
Maybe we should discuss with @innnkkki if they are interested in contributing to the docs.

PR #268, opened 2021-10-05
[PROD-5678] Validate visitor cookie before adding it to state

klesgidis:
## Summary
* Validates that visitor is a uuid before adding it to state
* Converts `error` to `warn` when failed to parse cookie


manoskouvarakis (approved):

nikostoulas (commented):
As discussed offline this can be disabled for projects that don't need it through
config.visitor.enabled = false.

We have pending to add the cookie setter in orka too.

PR #269, opened 2021-10-06
ENG-97: Expose more detailed bull metrics in order to distinguish wai…

kkyriakidis:
…ting jobs from delayed

https://workable.atlassian.net/browse/ENG-97

Kept the original bull naming for the jobs statuses


kkyriakidis:
> Nice @kkyriakidis 🎉 ! Is it your first pr?

Yes. First blood!

PR #276, opened 2022-01-03
Bump honeybadger library. Modify custom fingerprint

manoskouvarakis:
There are many occasions where current fingerprint groups errors that are not similar.
This PR should improve grouping by also using file and line to custom fingerprint.

### DIFF
--- a/src/initializers/honeybadger.ts
+++ b/src/initializers/honeybadger.ts
@@ -1,9 +1,19 @@
-export default async (config) => {
+export default async config => {
   if (config.honeybadger.apiKey) {
-    const honeybadger = await import('honeybadger');
+    const honeybadger = await import('@honeybadger-io/js');
     honeybadger.configure({
+      apiKey: config.honeybadger.apiKey,
       environment: config.app.env,
       developmentEnvironments: config.honeybadger.developmentEnvironments
+    });
+    honeybadger.beforeNotify(function (notice) {
+      // For testing purposes. Will delete after
+      console.log(`fingerprint before: ${notice.fingerprint}`);
+      
+      if (notice.backtrace.length > 0) {
+        notice.fingerprint += `_${notice.backtrace[0].file}_${notice.backtrace[0].number}`;
+      }
+      console.log(`fingerprint after: ${notice.fingerprint}`);
     });
   }
 };
--- a/package-lock.json
+++ b/package-lock.json
@@ -431,6 +431,38 @@
         "@hapi/hoek": "^9.0.0"
       }
     },
+    "@honeybadger-io/js": {
+      "version": "3.2.7",
+      "resolved": "https://registry.npmjs.org/@honeybadger-io/js/-/js-3.2.7.tgz",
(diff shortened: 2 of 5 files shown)


src/initializers/honeybadger.ts
```diff
@@ -1,9 +1,19 @@
...
+export default async config => {
   if (config.honeybadger.apiKey) {
-    const honeybadger = await import('honeybadger');
+    const honeybadger = await import('@honeybadger-io/js');
     honeybadger.configure({
+      apiKey: config.honeybadger.apiKey,
       environment: config.app.env,
       developmentEnvironments: config.honeybadger.developmentEnvironments
     });
+    honeybadger.beforeNotify(function (notice) {
```
nikostoulas, line 9: 🔧  We set the fingerprint on honeybadger-appender.ts. Please move the logic there.

manoskouvarakis: I added it there because it seems like the proposed place from [hb documentation](https://docs.honeybadger.io/lib/javascript/guides/customizing-error-grouping/).
Backtrace also exists on callback, not in appender. We could move all fingerprint calculation, but I preferred to have it calculated on appender and "enhance" it on callback

nikostoulas: I would put it in one place (appender) and use the error and error.stacktrace to enhance it. (I guess this is what backtrace does internally but this way you won't have to read the lib's code to understand what is being logged and if it is nullable etc.)

nikostoulas: I see that honeybadger-js is using https://www.npmjs.com/package/stack-trace so the way you do it is the easier one without also importing this lib. Just remove the console.logs then and leave it as is.

nikostoulas:
> There are many occasions where current fingerprint groups errors that are not similar. This PR should improve grouping by also using file and line to custom fingerprint.
> 
> Fingerprint before this PR: **rabbit-queue_parse**
> 
> Fingerprint after this PR: **rabbit-queue_parse_[PROJECT_ROOT]/src/app/services/flows/ParseFlow.ts_74**

We should use this in one app in officedroid before merging this to see if the new fingerprint is fitting our needs. We can install orka directly from github. Sth like ```npm install use_name/node_project#branch_name```

manoskouvarakis:
> > There are many occasions where current fingerprint groups errors that are not similar. This PR should improve grouping by also using file and line to custom fingerprint.
> > Fingerprint before this PR: **rabbit-queue_parse**
> > Fingerprint after this PR: **rabbit-queue_parse_[PROJECT_ROOT]/src/app/services/flows/ParseFlow.ts_74**
> 
> We should use this in one app in officedroid before merging this to see if the new fingerprint is fitting our needs. We can install orka directly from github. Sth like `npm install use_name/node_project#branch_name`

I have already done it to maestro officedroid. Let's use it for some days because errors are few until now to be sure if we prefer it

src/initializers/honeybadger.ts
manoskouvarakis: console.logs are removed

nikostoulas: Checked the maestro a bit. I am worried we will keep having new error groups for known errors, e.g. check https://app.honeybadger.io/projects/50040/faults?q=component%3Arabbit-queue+action%3Aparse+environment%3Aofficedroid&sort=last_seen_desc.
With every deploy that changes the line of an error we will get a different group.

PR #277, opened 2022-01-13
Adds default middleware to parse querystring

klesgidis:
## Summary
Adds default middleware to parse querystring.

Converts: `firstname=john&lastname=doe&children[]=bolek&children[]=lolek`
to 
```
ctx.state.query = {
    firstname: 'john',
    lastname: 'doe',
    children: ['bolek', 'lolek']
};
```


src/initializers/koa/parse-querystring.ts
```diff
@@ -0,0 +1,9 @@
+import { Context, Next } from 'koa';
+import * as qs from 'qs';
+
+export default async function parseQuerystring(ctx: Context, next: Next) {
+  if (ctx.querystring) {
+    ctx.state.query = qs.parse(ctx.querystring);
```
nikostoulas, line 6: I am not sure how performant qs.parse is. But I would add it optionally. E.g. only when someone accesses ctx.state.query

nikostoulas (approved):

nikosd23 (approved):

src/initializers/koa/parse-querystring.ts
klesgidis: How could you know that someone is accessing `ctx.state.query`?

nikostoulas: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty

src/initializers/koa/parse-querystring.ts
```diff
@@ -0,0 +1,13 @@
+import { Context, Next } from 'koa';
+import * as qs from 'qs';
+
+export default async function parseQuerystring(ctx: Context, next: Next) {
+  if (ctx.querystring) {
+    Object.defineProperty(ctx.state, 'query', {
+      get() {
```
nikostoulas, line 7: Note that ctx.state.query will be immutable since no setter is defined!

nikostoulas (approved):

jgerbe (approved):

PR #283, opened 2022-02-28
dd logs and metrics for kafka consumers

nikostoulas:
kafka_logs

### DIFF
--- a/src/initializers/kafka/base-kafka-handler.ts
+++ b/src/initializers/kafka/base-kafka-handler.ts
@@ -4,6 +4,7 @@ import { flatten } from 'lodash';
 import type * as KafkajsType from 'kafkajs';
 import { runWithContext } from '../../builder';
 import { alsSupported } from '../../utils';
+import { logMetrics } from '../../helpers';
 
 export type BaseKafkaHandlerOptions = {
   topic: string | string[];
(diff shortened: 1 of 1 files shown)


manoskouvarakis (approved):

src/initializers/kafka/base-kafka-handler.ts
```diff
@@ -73,10 +74,7 @@ abstract class Base<Input, Output> {
   }
 
   parseHeaders(headers: KafkajsType.IHeaders) {
-    return Object.keys(headers).reduce(
-      (m, k) => ({ ...m, [k]: headers[k].toString() }),
-      {}
-    );
+    return Object.keys(headers).reduce((m, k) => ({ ...m, [k]: headers[k].toString() }), {});
```
klesgidis, line 77: nit: 
```suggestion
    return Object.entries(headers).reduce((m, [k, v]) => ({ ...m, [k]: v.toString() }), {});
```

klesgidis (approved):

src/initializers/kafka/base-kafka-handler.ts
klesgidis: Actually I this is not supported in Node 10, so lets leave it as you did it

nikostoulas: This is prettier doing the auto-formatting. Didn't touch those lines. (I am wondering why prettier didn't work for you)

PR #303, opened 2022-07-26
PROD-13615 - Log warning for errors with only warn flag

Dimitris-Ilias:
https://workable.atlassian.net/browse/PROD-13615

### DIFF
--- a/src/initializers/koa/error-handler.ts
+++ b/src/initializers/koa/error-handler.ts
@@ -2,7 +2,7 @@ import * as Koa from 'koa';
 import { getLogger } from '../log4js';
 import { OrkaOptions } from 'orka/typings/orka';
 import { omit } from 'lodash';
-const Levels = require('log4js/lib/levels');
+import { levels as Levels } from 'log4js';
 const logger = getLogger('orka.errorHandler');
 
 export const isBlacklisted = (err: { status: number } = {} as any, config) =>
--- a/src/initializers/koa/error-handler.ts
+++ b/src/initializers/koa/error-handler.ts
@@ -2,12 +2,20 @@ import * as Koa from 'koa';
 import { getLogger } from '../log4js';
 import { OrkaOptions } from 'orka/typings/orka';
 import { omit } from 'lodash';
-
+const Levels = require('log4js/lib/levels');
 const logger = getLogger('orka.errorHandler');
 
 export const isBlacklisted = (err: { status: number } = {} as any, config) =>
   // tslint:disable-next-line:triple-equals
   err.status && config.blacklistedErrorCodes.some(b => err.status == b);
+
+export const onlyWarn = (err) => {
+  const levelStr = err?.logLevel;
+  if (typeof levelStr !== 'string') return false;
+  const level = Levels[levelStr.toUpperCase()]?.level;
+  if (!level) return false;
+  return level <= Levels.WARN.level;
+};
 
 export default (config, orkaOptions: Partial<OrkaOptions>) =>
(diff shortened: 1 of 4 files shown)


Dimitris-Ilias:
@nikostoulas Wdyt? Can't we can do sth like this to hide indeed prediction/recommendation errors?
Didn't want to use the blacklist because this could potentially hide errors of other job boards

src/initializers/koa/error-handler.ts
```diff
@@ -9,6 +9,8 @@ export const isBlacklisted = (err: { status: number } = {} as any, config) =>
   // tslint:disable-next-line:triple-equals
   err.status && config.blacklistedErrorCodes.some(b => err.status == b);
 
+const onlyWarn = (err: { onlyWarn: boolean }) => !!err?.onlyWarn
```
nikostoulas, line 12:
```suggestion
const logLevel = (err: { logLevel: LogLevels }) => !!err?.logLevel
```

nikostoulas (commented):
Yeah we could do sth like this @Dimitris-Ilias

src/initializers/koa/error-handler.ts
Dimitris-Ilias: @nikostoulas Can you take a look at the latest implementation?

I was checking https://www.npmjs.com/package/http-errors documentation
image


So based on the `properties` field, in order to make this work with `ctx.assert` maybe only thing that's needed is to use it like this on the postings server side?
`ctx.assert(token, 401, { message: calculateMessage(), logLevel: 'WARN' });`

src/initializers/koa/error-handler.ts
```diff
@@ -33,10 +41,10 @@ export default (config, orkaOptions: Partial<OrkaOptions>) =>
...
 
       const errorArgs = (await orkaOptions.errorHandler(ctx, err, orkaOptions)) || [err as Error];
-      if (!isBlacklisted(err, config)) {
-        logger.error(...errorArgs);
-      } else {
+      if (isBlacklisted(err, config) || onlyWarn(err)) {
         logger.warn(...errorArgs);
+      } else {
+        logger.error(...errorArgs);
       }
```
nikostoulas, line 48:
```suggestion
   logger[getLogLevel(err, config)](...errorArgs)
   
   and    getLoglevel sth like
   
   - if err.logLevel is given and is valid use this
   - else if blacklisted warn
   - else error
   
```

src/initializers/koa/error-handler.ts
nikostoulas: Nice interface @Dimitris-Ilias .
Left a comment for the implementation details

src/initializers/koa/error-handler.ts
Dimitris-Ilias: @nikostoulas Done, cytal?

src/initializers/koa/error-handler.ts
```diff
@@ -2,13 +2,21 @@ import * as Koa from 'koa';
 import { getLogger } from '../log4js';
 import { OrkaOptions } from 'orka/typings/orka';
 import { omit } from 'lodash';
-
+const Levels = require('log4js/lib/levels');
```
nikostoulas, line 5: 🔧 can we convert this to import?

nikostoulas (approved):
Nice @Dimitris-Ilias. Since we are touching error handler can we document it's behavior here: https://workable.github.io/orka/middleware.html ?

src/initializers/koa/error-handler.ts
Dimitris-Ilias: Done

PR #304, opened 2022-08-22
Joi fixes & riviere default config override

klesgidis:
## Summary
* joi: Remove buggy allowEmpty rule
  * Someone can replace with `.allow('null', '')`
* riviere: override defaults inbound/outbound if needed

### DIFF
--- a/src/initializers/diamorphosis.ts
+++ b/src/initializers/diamorphosis.ts
@@ -68,14 +68,18 @@ export default (config, orkaOptions: Partial<OrkaOptions>) => {
     enabled: true,
     inbound: {
       request: {
-        enabled: false
-      }
+        enabled: false,
+        ...config.riviere?.inbound?.request
+      },
+      ...config.riviere?.inbound
     },
     outbound: {
       blacklistedPathRegex: new RegExp('^/v0.4/traces$', 'i'),
       request: {
-        enabled: false
-      }
+        enabled: false,
+        ...config.riviere?.outbound?.request
+      },
+      ...config.riviere?.outbound
     },
     color: true,
     styles: [],
--- a/src/initializers/diamorphosis.ts
+++ b/src/initializers/diamorphosis.ts
@@ -67,19 +67,19 @@ export default (config, orkaOptions: Partial<OrkaOptions>) => {
   config.riviere = {
     enabled: true,
     inbound: {
+      ...config.riviere?.inbound,
       request: {
         enabled: false,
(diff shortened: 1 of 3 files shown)


src/initializers/diamorphosis.ts
```diff
@@ -68,14 +68,18 @@ export default (config, orkaOptions: Partial<OrkaOptions>) => {
     enabled: true,
     inbound: {
       request: {
-        enabled: false
-      }
+        enabled: false,
+        ...config.riviere?.inbound?.request
+      },
+      ...config.riviere?.inbound
```
nikostoulas, line 74: ❓ @klesgidis in order to not overwrite request.enabled:false this should be on the top right?

src/initializers/diamorphosis.ts
```diff
@@ -68,14 +68,18 @@ export default (config, orkaOptions: Partial<OrkaOptions>) => {
...
     },
     outbound: {
       blacklistedPathRegex: new RegExp('^/v0.4/traces$', 'i'),
       request: {
-        enabled: false
-      }
+        enabled: false,
+        ...config.riviere?.outbound?.request
+      },
+      ...config.riviere?.outbound
```
nikostoulas, line 82: same

test/initializers/joi.test.ts
```diff
@@ -140,7 +140,7 @@ describe('joi extensions', function () {
         '<a href="/j/ABCDEFG">banana</a>'
       );
     });
-    it('invalid tags', function() {
+    it.only('invalid tags', function() {
```
nikostoulas, line 143: rm only

src/initializers/joi.ts
```diff
@@ -171,14 +171,6 @@ const Joi: JoiWithExtensions = _Joi.extend(
   joi => ({
     type: 'safeHtml',
     base: joi.string(),
-    rules: {
-      allowEmpty: {
-        method() {
-          // @ts-ignore
-          return this.$_root.allow('').allow(null);
```
nikostoulas, line 178: ❓ Why was this buggy?

klesgidis: This overrides the whole scema, an the `prepare` method is not called.

In addition, `allowEmpty` can be achieved with `.allow('', null)`

nikostoulas (approved):

src/initializers/joi.ts
nikostoulas: Thanks for the explanation

PR #309, opened 2022-10-12
Add changelog notes for migrating to Orka 3.*

kkyriakidis:


nikostoulas (approved):
Thank you @kkyriakidis . Are any steps necessary to migrate an app to latest orka that we could also mention? E.g. maybe a mongodb version compatibility check?

klesgidis (approved):

kkyriakidis:
The mongoDB Node.js driver derives from the mongoose version. Unless specific stated. There can be incompatibilities with libraries using the driver. In our case in assessments we had an issue with ts-mongoose for example. 

Also we had breaking changes noted to the mongoose migration guide 
Types.ObjectId is now a class, which means you can no longer omit new when creating a new ObjectId using new mongoose.Types.ObjectId(). Currently, you can still omit new in JavaScript, but you must put new in TypeScript.
Do you think I should specific add this to the orka changelog? @nikostoulas

nikostoulas:
> The mongoDB Node.js driver derives from the mongoose version. Unless specific stated. There can be incompatibilities with libraries using the driver. In our case in assessments we had an issue with ts-mongoose for example.
> 
> Also we had breaking changes noted to the mongoose migration guide Types.ObjectId is now a class, which means you can no longer omit new when creating a new ObjectId using new mongoose.Types.ObjectId(). Currently, you can still omit new in JavaScript, but you must put new in TypeScript. Do you think I should specific add this to the orka changelog? @nikostoulas

If you think they are common sure. If not then link to the docs is ok.

nikostoulas (approved):
⭐

PR #313, opened 2022-11-24
Do not create topics if already exist

klesgidis:
## Summary
Do not create topics if already exist


nikostoulas (approved):
@klesgidis we use createTopics in production without any issues. What was the rational behind this pr? Did it cause any issues or is it to have more accurate logs?

klesgidis:
> @klesgidis we use createTopics in production without any issues. What was the rational behind this pr? Did it cause any issues or is it to have more accurate logs?

I could see the errors be logged. and it is also sends error to HB

PR #315, opened 2022-12-28
Teach orka to getRootSpan for all versions of dd-trace

manoskouvarakis:
Currently orka (and all our apps) work with dd-trace until version 2.2.0.

This is because dd-trace library deprecated (private) property req._datadog we use ([example](https://github.com/Workable/orka/blob/c915badf5793da3db516c1d5a2f98d0e1d6ca816/src/middlewares/datadog-match-routes.ts#L5)) to get the root span.

Here are the respective changes between versions [2.2.0](https://github.com/DataDog/dd-trace-js/blob/4a096167f4b5168186df9324f4a4a1364ce12f87/packages/dd-trace/src/plugins/util/web.js#L161) and [2.2.1](https://github.com/DataDog/dd-trace-js/blob/a331be34e1ed3b2b8461a4993e425e107f12664a/packages/dd-trace/src/plugins/util/web.js#L169) in dd-trace that caused our implementation to stop working if we bump (tested in prisma)

SRE team asked for a version > 2.15.0 that supports an environment variable they want to use. 

After a lot of testing, the only workaround found is based on this [issue](https://github.com/DataDog/dd-trace-js/issues/725) that asks for the exact same span we need (fun fact is that it is [implemented in helper method of dd-trace](https://github.com/DataDog/dd-trace-js/blob/b78ebaaea3e4857b4de3cd4112d0d9ed2c2f6730/packages/dd-trace/src/plugins/util/web.js#L251) library but unfortunately not exposed in api)

New orka helper is tested in prisma with dd-trace versions 1.7.1, 2.20.0, 3.9.3 and seems to work as expected. In the screenshot above there are requests with version 3.9.3 (at 14:17) and 1.7.1 (at 14:44). [Here](https://app.datadoghq.com/apm/traces?query=%40_top_level%3A1%20env%3Aofficedroid%20service%3Aprisma%20operation_name%3Akoa.request&cols=core_service%2Ccore_resource_name%2Clog_duration%2Clog_http.method%2Clog_http.status_code&historicalData=true&messageDisplay=inline&sort=desc&spanType=service-entry&start=1672228628846&end=1672232228846&paused=false) is the link of DD (it traces are still available)

image

image

Before the fix, even the url wasn't available in resource.

### DIFF
--- a/src/helpers/get-root-span.ts
+++ b/src/helpers/get-root-span.ts
@@ -1,6 +1,8 @@
-import { getDatadogTracer } from '../initializers/datadog/index';
+import { getDatadogTracer, isDatadogEnabled } from '../initializers/datadog/index';
 
 export default function getRootSpan(ctx) {
+  if (!isDatadogEnabled()) return;
+
   let ddSpan = ctx.req?._datadog?.span;
   if (!ddSpan) {
     const activeSpan = getDatadogTracer()?.scope()?.active();
--- a/src/helpers/get-root-span.ts
+++ b/src/helpers/get-root-span.ts
@@ -1,0 +1,10 @@
+import { getDatadogTracer } from '../initializers/datadog/index';
+
+export default function getRootSpan(ctx) {
+  let ddSpan = ctx.req?._datadog?.span;
+  if (!ddSpan) {
+    const activeSpan = getDatadogTracer().scope().active();
+    if (activeSpan) ddSpan = activeSpan.context()._trace.started[0];
+  }
+  return ddSpan;
+}
--- a/src/helpers/get-root-span.ts
+++ b/src/helpers/get-root-span.ts
@@ -1,12 +1,19 @@
 import { getDatadogTracer, isDatadogEnabled } from '../initializers/datadog/index';
+import { getLogger } from '../initializers/log4js';
+
+const logger = getLogger('orka.helpers.get-root-span');
 
 export default function getRootSpan(ctx) {
   if (!isDatadogEnabled()) return;
 
   let ddSpan = ctx.req?._datadog?.span;
   if (!ddSpan) {
-    const activeSpan = getDatadogTracer()?.scope()?.active();
-    if (activeSpan?.context()?._trace?.started.length > 0) ddSpan = activeSpan?.context()?._trace?.started[0];
+    try {
+      const activeSpan = getDatadogTracer()?.scope()?.active();
+      if (activeSpan?.context()?._trace?.started.length > 0) ddSpan = activeSpan?.context()?._trace?.started[0];
+    } catch (e) {
+      logger.error('dd-trace error trying to find root span', e);
+    }
   }
   return ddSpan;
 }
--- a/src/helpers/get-root-span.ts
+++ b/src/helpers/get-root-span.ts
@@ -3,8 +3,8 @@ import { getDatadogTracer } from '../initializers/datadog/index';
 export default function getRootSpan(ctx) {
   let ddSpan = ctx.req?._datadog?.span;
   if (!ddSpan) {
-    const activeSpan = getDatadogTracer().scope().active();
-    if (activeSpan) ddSpan = activeSpan.context()._trace.started[0];
+    const activeSpan = getDatadogTracer()?.scope()?.active();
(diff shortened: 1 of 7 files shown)


src/helpers/get-root-span.ts
```diff
@@ -0,0 +1,12 @@
+import { getDatadogTracer, isDatadogEnabled } from '../initializers/datadog/index';
+
+export default function getRootSpan(ctx) {
+  if (!isDatadogEnabled()) return;
+
+  let ddSpan = ctx.req?._datadog?.span;
+  if (!ddSpan) {
+    const activeSpan = getDatadogTracer()?.scope()?.active();
```
nikostoulas, line 8: ❓ Could this be throwing errors in a future dd-trace that does not implement scope or active? Shall we try catch it?

src/helpers/index.ts
```diff
@@ -1,9 +1,6 @@
...
 import axiosErrorInterceptor from './interceptors/axios-error-interceptor';
 import getLock from './get-lock';
+import getRootSpan from './get-root-span';
 
-export {
-  logMetrics,
-  axiosErrorInterceptor,
-  getLock
-};
+export { logMetrics, axiosErrorInterceptor, getLock, getRootSpan };
```
nikostoulas, line 6: Nice! Shall we document getRootSpan in our public docs? Or do we prefer this being kept "internal" ?

src/initializers/datadog/index.ts
```diff
@@ -19,3 +19,7 @@ export const getDatadogTracer = () => {
   }
   return tracer;
 };
+
+export const isDatadogEnabled = () => {
```
nikostoulas, line 23: 👏 very clear!

test/helpers/get-root-span.test.ts
```diff
@@ -0,0 +1,74 @@
...
+
+      const ctx = ({ request: { method: 'GET' } } as unknown) as Context;
+      const span = getRootSpan(ctx);
+      should(span).be.undefined();
+    });
+
+    it('should return span from started array first value', function () {
+      const tracerStub = {
+        scope: sandbox.stub().returns({
+          active: sandbox.stub().returns({
```
nikostoulas, line 62: ❓ Shall we add a test where active is not defined for example? Is this documented in dd-trace and expected to be exposed in future versions too?

nikostoulas (approved):
Excellent job @manoskouvarakis. Very nice and detailed description and good job with testing different versions before opening the pr!
Keep it up 👏 
Some minor comments to make it future proof and possibly expose the functionality to documentation.

test/helpers/get-root-span.test.ts
manoskouvarakis: Yes it is documented [here](https://datadoghq.dev/dd-trace-js/interfaces/scope.html) so if it is deprecated we will see it in change log.

npetridis (approved):
Great job!

src/helpers/get-root-span.ts
manoskouvarakis: Added try catch and log error if anything breaks

src/helpers/index.ts
manoskouvarakis: I believe we can keep it internal for now, until either we use it in all our apps and everything works as expected or it is documented in dd-trace lib. Wdyt?

src/helpers/get-root-span.ts
```diff
@@ -0,0 +1,19 @@
...
+export default function getRootSpan(ctx) {
+  if (!isDatadogEnabled()) return;
+
+  let ddSpan = ctx.req?._datadog?.span;
+  if (!ddSpan) {
+    try {
+      const activeSpan = getDatadogTracer()?.scope()?.active();
+      if (activeSpan?.context()?._trace?.started.length > 0) ddSpan = activeSpan?.context()?._trace?.started[0];
+    } catch (e) {
+      logger.error('dd-trace error trying to find root span', e);
```
nikostoulas, line 15:
```suggestion
      logger.error(e, 'dd-trace error trying to find root span');
```

nikostoulas: If the first argument is not an error it creates a new error with the string given here: https://github.com/Workable/orka/blob/master/src/initializers/log4js/honeybadger-appender.ts#L7

manoskouvarakis: Thanks! Fixed

PR #321, opened 2023-02-22
Set logging in metrics optional

manoskouvarakis:
The purpose of this PR is to allow each application using logMetrics to define if it will log TIME_LOGGING message apart from sending it to DD.

### DIFF
--- a/src/helpers/log-metrics.ts
+++ b/src/helpers/log-metrics.ts
@@ -22,11 +22,14 @@ const logMetrics = {
     const s = Number(ns) / 1e9;
     if (prometheusTimeEnabled()) logMetrics.prometheusEndClient().observe({ flow, flowType }, s);
     if (newRelicEnabled()) getNewRelic().recordMetric(`Custom/${flowType}/${flow}`, s);
+
     if (enableLog) logger.debug(`[${id}] TIME_LOGGING[${flowType}][${flow}] ${s.toFixed(3)} s`);
+    else logger.trace(`[${id}] TIME_LOGGING[${flowType}][${flow}] ${s.toFixed(3)} s`);
   },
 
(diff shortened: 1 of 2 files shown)


nikostoulas (approved):

src/helpers/log-metrics.ts
```diff
@@ -17,16 +17,19 @@ const logMetrics = {
...
-  end(start: bigint, flow: string, flowType: string, id: string) {
+  end(start: bigint, flow: string, flowType: string, id: string, enableLog: boolean = true) {
     const ns = process.hrtime.bigint() - start;
     const s = Number(ns) / 1e9;
     if (prometheusTimeEnabled()) logMetrics.prometheusEndClient().observe({ flow, flowType }, s);
     if (newRelicEnabled()) getNewRelic().recordMetric(`Custom/${flowType}/${flow}`, s);
-    logger.debug(`[${id}] TIME_LOGGING[${flowType}][${flow}] ${s.toFixed(3)} s`);
+
+    if (enableLog) logger.debug(`[${id}] TIME_LOGGING[${flowType}][${flow}] ${s.toFixed(3)} s`);
+    else logger.trace(`[${id}] TIME_LOGGING[${flowType}][${flow}] ${s.toFixed(3)} s`);
```
nikostoulas, line 27: 🔧 we should update the param variable to determine the log level instead.

manoskouvarakis: Done. I preferred to be a boolean again as it is easier for those two levels

manoskouvarakis:
This is probably not relevant any more as [we will not index](https://workable.atlassian.net/wiki/spaces/SRE/pages/4076535827/Datadog+Logging+Migration+Requirements) DEBUG logs in Datadog (only ingest). Closing for now

PR #322, opened 2023-04-12
Do not require mongodb before orka is initialized

nikostoulas:
Requiring mongodb before orka is initialized does not work well with datadog, newrelic.

### DIFF
--- a/.projections.json
+++ b/.projections.json
@@ -1,0 +1,10 @@
+{
+  "src/*.ts": {
+    "alternate": "test/{}.test.ts",
+    "type": "test"
+  },
+  "test/*.test.ts": {
+    "alternate": "src/{}.ts",
+    "type": "source"
+  }
+}
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -42,7 +42,7 @@ export default (defaults: Partial<OrkaOptions> = _defaults) => {
     logger.error(
       new Error(
         `Config in path ${options.diamorphosis.configPath} was already required. ` +
-        `Your config might be used before being initialized by orka (diamorphosis).`
+          `Your config might be used before being initialized by orka (diamorphosis).`
       )
     );
   }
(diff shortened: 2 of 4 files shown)


src/initializers/joi.ts
```diff
@@ -35,12 +34,12 @@ const Joi: JoiWithExtensions = _Joi.extend(
     base: joi.string().meta({ baseType: 'string' }),
     messages: { 'objectId.invalid': 'Invalid objectId' },
     validate: (value, helpers) => {
-      return mongodb.ObjectId.isValid(value)
+      return require('mongodb').ObjectId.isValid(value)
```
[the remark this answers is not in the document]
klesgidis: We don't use it at all in HRIS. I think it is used on careers or WJB

.projections.json
```diff
@@ -0,0 +1,10 @@
+{
```
klesgidis, line 1: what is this?

klesgidis (approved):

.projections.json
nikostoulas: https://marketplace.visualstudio.com/items?itemName=will-wow.vscode-alternate-file
or https://github.com/tpope/vim-projectionist
I have a shortcut for moving between implementation and test files and thought it might be useful for others too.
(My shortcut is <space>gt  (go to test))

src/initializers/joi.ts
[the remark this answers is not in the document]
nikostoulas: Careers page uses it's own vailidation not orka's. So this is probably not used anywhere in our codebase.

src/builder.ts
```diff
@@ -52,10 +52,16 @@ export default (defaults: Partial<OrkaOptions> = _defaults) => {
   diamorphosis(config, options);
 
   options.appName ||= config?.app?.name;
-  if (require.cache[require.resolve('koa')]) logger.warn('Koa was initialized before orka');
-  if (config.mongodb?.url && require.cache[require.resolve('mongoose')]) logger.warn('Mongoose was initialized before orka');
-  if (config.queue?.url && require.cache[require.resolve('amqplib')]) logger.warn('Amqplib was initialized before orka');
-  if (config.postgres?.url && require.cache[require.resolve('pg')]) logger.warn('postgres was initialized before orka');
```
manoskouvarakis, line 58: ❓ Don't we also need the condition for the config part? Why it can be omitted?

nikostoulas: The condition was needed because pg for example might not be installed and require would throw an error. The error is caught in the new implementation

PR #325, opened 2023-05-24
Propagate white-listed headers on  request context to subsequent calls

Michalis-Apostolou:
In order to propagate 'cf-ray' header to subsequent calls as requested here https://workable.atlassian.net/browse/PROD-20907

provided a generalised functionality for each application to define which headers can be propagated for logging.


manoskouvarakis (approved):
👏

nikostoulas (approved):

PR #331, opened 2023-08-29
[PROD-26633] Introduce growthbook to orka

klesgidis:
## Summary
* Adds growthbook initialization
* Adds growthbook middleware
* Adds `getGrowthbook` method to get the instance (supports generic typing for auto completion)
* Adds tests
* Adds documentation

### DIFF
--- a/src/initializers/growthbook.ts
+++ b/src/initializers/growthbook.ts
@@ -1,0 +1,22 @@
+import requireInjected from '../require-injected';
+import type * as GrowthbookType from 'growthbook';
+
+let gb: GrowthbookType.GrowthBook;
+
+export default config => {
+  if (!gb) {
+    const {GrowthBook} = requireInjected('growthbook');
+    gb = new GrowthBook({
+      apiHost: 'https://cdn.growthbook.io',
+      clientKey: config.growthbook.clientKey
+    });
+  }
+  return gb;
+};
+
+export const getGrowthbook = () => {
+  if (!gb) {
+    throw new Error('growthbook is not initialized');
+  }
+  return gb;
+};
--- a/src/initializers/growthbook.ts
+++ b/src/initializers/growthbook.ts
@@ -3,10 +3,11 @@ import type * as GrowthbookType from 'growthbook';
 
 let gb: GrowthbookType.GrowthBook;
 
-export default config => {
+const createGrowthbook = <AppFeatures extends Record<string, any> = Record<string, any>>(config) => {
+  if (!config.growthbook) return;
   if (!gb) {
     const {GrowthBook} = requireInjected('growthbook');
-    gb = new GrowthBook({
+    gb = new GrowthBook<AppFeatures>({
       apiHost: 'https://cdn.growthbook.io',
       clientKey: config.growthbook.clientKey
     });
@@ -14,9 +15,11 @@ export default config => {
   return gb;
 };
 
-export const getGrowthbook = () => {
+export default createGrowthbook;
+
+export const getGrowthbook = <AppFeatures extends Record<string, any> = Record<string, any>>() => {
   if (!gb) {
     throw new Error('growthbook is not initialized');
   }
-  return gb;
+  return gb as GrowthbookType.GrowthBook<AppFeatures>;
 };
--- a/src/middlewares/growthbook.ts
+++ b/src/middlewares/growthbook.ts
@@ -1,0 +1,17 @@
+import {Context} from 'koa';
+import {getGrowthbook, getRequestContext} from '../index';
+import {getLogger} from '../initializers/log4js';
+
+const logger = getLogger('orka.middlewares.growthbook');
+
+export default async function growthbook(ctx: Context, next: () => Promise<void>) {
+  if (ctx.path === '/health') return next();
+  const gb = getGrowthbook();
+  if (!gb) return next();
+  ctx.state.growthbook = gb;
+  await gb.loadFeatures().catch(e => logger.error(e, 'failed to load growthbook'));
+  if (ctx.state.growthbook) {
+    getRequestContext()?.set('growthbook', ctx.state.growthbook);
+  }
+  await next();
+}
--- a/src/orka-builder.ts
+++ b/src/orka-builder.ts
@@ -29,6 +29,7 @@ import * as Koa from 'koa';
 import { AsyncLocalStorage } from 'async_hooks';
 import { alsSupported } from './utils';
 import type WorkerType from './initializers/worker';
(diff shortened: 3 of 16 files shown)


jgerbe (approved):
👏

src/middlewares/growthbook.ts
```diff
@@ -0,0 +1,17 @@
...
+import {getLogger} from '../initializers/log4js';
+
+const logger = getLogger('orka.middlewares.growthbook');
+
+export default async function growthbook(ctx: Context, next: () => Promise<void>) {
+  if (ctx.path === '/health') return next();
+  const gb = getGrowthbook();
+  if (!gb) return next();
+  ctx.state.growthbook = gb;
+  await gb.loadFeatures().catch(e => logger.error(e, 'failed to load growthbook'));
```
nikostoulas, line 12: 🔧 This adds a delay to each and every http call. 
Removed await in job-board for this reason, features might be delayed a few ms the first time a service starts up

src/initializers/growthbook.ts
```diff
@@ -0,0 +1,25 @@
...
+      apiHost: 'https://cdn.growthbook.io',
+      clientKey: config.growthbook.clientKey
+    });
+  }
+  return gb;
+};
+
+export default createGrowthbook;
+
+export const getGrowthbook = <AppFeatures extends Record<string, any> = Record<string, any>>() => {
```
nikostoulas, line 20: 🔧 Per growthbooks documentation you create and destroy an instance every time https://docs.growthbook.io/lib/js.
If you don't there is a risk of mixing features (e.g. an account with a FF enabled might interfere with what happens to one that has the same FF disabled)

nikostoulas (commented):
This PR does not provide a paradigm for `setAttributes`. This is important as Growthbook's value is the easy enablement of FF for specific part of the userbase, gradual rollouts etc.

src/middlewares/growthbook.ts
klesgidis: I know. I saw the comments, but I don't think it is orka's decision to not await this call. Some apps may need to have features loaded from the first call. 

In any case, this caches the requests to growthbook correct?

src/initializers/growthbook.ts
klesgidis: Ok, I'll address

src/initializers/growthbook.ts
```diff
@@ -0,0 +1,22 @@
+import requireInjected from '../require-injected';
+import type * as GrowthbookType from 'growthbook';
+
+let gb: GrowthbookType.GrowthBook;
+
+export default config => {
+  if (!gb) {
+    const {GrowthBook} = requireInjected('growthbook');
+    gb = new GrowthBook({
+      apiHost: 'https://cdn.growthbook.io',
```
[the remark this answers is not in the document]
klesgidis: I'll address

src/initializers/growthbook.ts
klesgidis: The problem here though, is that we need to give the caller the responsibility to destroy the instance after they use `getGrowthbook`

klesgidis: @nikostoulas any idea if we could somehow handle this?

nikostoulas: See how we handle this in job-board https://github.com/Workable/job-board/blob/master/src/app/middlewares/growth-book.ts#L22

klesgidis:
> This PR does not provide a paradigm for `setAttributes`. This is important as Growthbook's value is the easy enablement of FF for specific part of the userbase, gradual rollouts etc.

I have a reference [here](https://github.com/Workable/orka/pull/331/files#diff-2e2b0cef59a054a44d89f4d8f52647832e1e0bd02c386fdc8ae9e34763874565R170-R171) for setAttributes. 

What do you mean a paradigm? I think it is in the caller's responsibility to setAttributes, no?

src/initializers/growthbook.ts
klesgidis: This is the easy part. What if someone calls `getGrowthbook` from a queue handler?

nikostoulas:
> > This PR does not provide a paradigm for `setAttributes`. This is important as Growthbook's value is the easy enablement of FF for specific part of the userbase, gradual rollouts etc.
> 
> I have a reference [here](https://github.com/Workable/orka/pull/331/files#diff-2e2b0cef59a054a44d89f4d8f52647832e1e0bd02c386fdc8ae9e34763874565R170-R171) for setAttributes.
> 
> What do you mean a paradigm? I think it is in the caller's responsibility to setAttributes, no?

No the way it is set up the caller will most probably end up setting attributes in parallel leading to users getting FF they should not. This should be solved by orka somehow by default and exclude the possibility of race conditions by design.

src/middlewares/growthbook.ts
nikostoulas: Not sure if it was only the first call however or all of them. Do you mind checking?

src/initializers/growthbook.ts
nikostoulas: You should not expose getGrowthbook. In this case it is the callers responsibility to call destroy.

klesgidis: I need to expose, in order to be able to use it on queues, consumers, workers etc

src/middlewares/growthbook.ts
klesgidis: I'll check

src/initializers/growthbook.ts
nikostoulas: I mean you should expose a createGrowthbook as growthbook instances should not be shared!

src/orka-builder.ts
```diff
@@ -167,8 +166,13 @@ export default class OrkaBuilder {
...
   }
 
-  withGrowthbook() {
-    this.queue.push(() => growthbook(this.config));
+  loadGrowthbookFeatures() {
+    this.queue.push(() => {
+      const { createGrowthbook } = require('./initializers/growthbook');
+      const gb = createGrowthbook(this.config.growthbook);
+      if (!gb) return;
+      gb.loadFeatures().catch(e => getLogger('orka.growthbook').error('Unable to load features', e));
```
nikostoulas, line 174: 🙃 You can use await here. It will delay startup by a few ms but even during startup you won't ever get a wrong FF. WDYT?

nikostoulas (approved):

src/initializers/growthbook.ts
klesgidis: addressed

PR #332, opened 2023-08-29
growthbook: Set attributes from callback

klesgidis:


src/middlewares/growthbook.ts
```diff
@@ -12,6 +12,10 @@ export default async function growthbook(ctx: Context, next: () => Promise<void>
   const gb = createGrowthbook(config);
   ctx.state.growthbook = gb;
   gb.loadFeatures().catch(e => logger.error(e, 'failed to load growthbook'));
+  gb.setAttributes({
+    ...gb.getAttributes(),
```
nikostoulas, line 16: ❓ is the get needed? Will it ever have values if set is not called before that?

src/initializers/diamorphosis.ts
```diff
@@ -275,6 +276,11 @@ function addGrowthbookConfig(config) {
   config.growthbook = {
     apiHost: 'https://cdn.growthbook.io',
     clientKey: '',
+    setAttributesCallback: setAttributesCallback,
```
nikostoulas, line 279: 🙃 I would move this to https://github.com/Workable/orka/blob/024156548971dee9d997dd403f2bd09b2e6b560e/src/default-options.ts#L41 because we don't have functions in config usually.
No strong opinion though.

nikostoulas (approved):

PR #336, opened 2023-09-20
[PROD-30208] - add check on s3 url for @ after domain to prevent security risks

CharisVrettos:


src/initializers/joi.ts
```diff
@@ -190,7 +190,8 @@ const Joi: JoiWithExtensions = _Joi.extend(
           }
         ],
         validate(value, helpers, {bucket}) {
-          if (!isOwnS3Path(bucket, value)) {
+          const { hostname, protocol, pathname} = new URL(value);
+          if (!isOwnS3Path(bucket, `${protocol}//${hostname}${pathname}`)) {
```
klesgidis, line 194: Clever! 💯  
I think we could also simplify it by doing
```suggestion
          const { origin } = new URL(value);
          if (!isOwnS3Path(bucket, origin)) {
```

klesgidis (approved):

src/initializers/joi.ts
CharisVrettos: As talked offline we need the path name also to make the validation because s3 can have the bucket in the path after domain

tomtoump:
@CharisVrettos Wouldn't it be better to update `isOwnS3Path`'s definition? This would also solve the same issue for cases like [this](https://github.com/Workable/job-board/blob/a92096da319ffe90de3bde21ca4ae39c5177da2e/src/app/services/helpers/file-url-helpers.ts#L3), that import the specific function.

CharisVrettos:
> @CharisVrettos Wouldn't it be better to update `isOwnS3Path`'s definition? This would also solve the same issue for cases like [this](https://github.com/Workable/job-board/blob/a92096da319ffe90de3bde21ca4ae39c5177da2e/src/app/services/helpers/file-url-helpers.ts#L3), that import the specific function.

Hello @tomtoump . Could you please explain a bit more what do you mean by `update isOwnS3Path` definition?

tomtoump:
I was trying to say that the url parsing could take place in the function's definition [here](https://github.com/Workable/orka/blob/681add6e3d091580375fc6acc1b9d5179545ac6e/src/initializers/joi.ts#L9), instead of inside the `validate()` function of the `urlInOwnS3` joi extension.

CharisVrettos:
Yes, I understood later what you meant. Thought I had done the change in `isOwnS3Path` and found it weird. My objection on this is that the `isOwnS3Path` is used to check whether the bucket belongs to the url so are we sure that we want to have an action there that is irrelevant to the function's purpose?

CharisVrettos:
Sure, will move the destructuring in `isOwnS3Path`

PR #338, opened 2023-09-27
Updates logs to be consistent - bumps versions

nikostoulas:
## Before notice the different last id

Screenshot 2023-09-27 at 6 27 56 PM

## After all ids are the same

Screenshot 2023-09-27 at 6 28 35 PM

Reproduce:
```
npx nodemon examples/request-context-example/app.js

...

curl -X GET localhost:2121/first -H 'x-request-id: bar' -H 'cf-ray: foo'
```

### DIFF
--- a/src/initializers/koa/add-request-id.ts
+++ b/src/initializers/koa/add-request-id.ts
@@ -1,8 +1,13 @@
 import { Context, Middleware } from 'koa';
+import { randomUUID } from 'crypto';
 
-export default function(config): Middleware {
+export default function (config): Middleware {
   return async function addRequestId(ctx: Context, next: () => void) {
-    const requestId = ctx.headers[config.traceHeaderName.toLowerCase()];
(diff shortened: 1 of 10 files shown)


src/initializers/koa/add-request-id.ts
```diff
@@ -1,8 +1,13 @@
 import { Context, Middleware } from 'koa';
+import { randomUUID } from 'crypto';
 
-export default function(config): Middleware {
+export default function (config): Middleware {
   return async function addRequestId(ctx: Context, next: () => void) {
-    const requestId = ctx.headers[config.traceHeaderName.toLowerCase()];
+    let requestId = ctx.headers[config.traceHeaderName.toLowerCase()];
+    if (!requestId) {
+      requestId = `orka-${randomUUID()}`;
```
nikostoulas, line 8: 🙃 If we don't set the requestId here riviere will set one for the server logs and a different one for the http requests. Setting it here make sure that the same id is shared everywhere.

PR #348, opened 2023-11-07
[PROD-32442] prevent ssrf attacks for own S3 urls

pmantafounis:
Prevent SSRF attacks in `isOwnS3Urls`.
Also replaced regex checking for s3 hosts with bucket in pathname, to be done using the `URL` attributes

### DIFF
--- a/src/initializers/joi.ts
+++ b/src/initializers/joi.ts
@@ -7,10 +7,18 @@ export const isValidPhone = (val: string): boolean => /^[\d\s\(\)\-\+–\.]+$/.t
 export const clearNullByte = (val: string): string => (val && isString(val) ? val.replace(/\u0000/g, '') : val);
 export const isValidHexColor = (val: string): boolean => /^#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3})$/.test(val);
 export const isOwnS3Path = (bucket: string, val: string): boolean => {
-  const { origin, pathname } = new URL(val);
-  val = `${origin}${pathname}`;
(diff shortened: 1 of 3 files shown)


src/initializers/joi.ts
```diff
@@ -7,10 +7,18 @@ export const isValidPhone = (val: string): boolean => /^[\d\s\(\)\-\+–\.]+$/.t
...
+    const matchingProtocol = protocol === 'https:';
+    const s3Host = host === `${bucket}.s3.amazonaws.com`;
+    const s3HostBucketInPath =
+      host.startsWith('s3.') &&
+      host.endsWith('.amazonaws.com') &&
+      pathname.startsWith(`/${bucket}/`);
+    return matchingProtocol && (s3Host || s3HostBucketInPath);
+  } catch (e) {
+    return false;
+  }
```
pmantafounis, line 21: A `logger.error` would be useful here, but was not sure if this is something we do in `orka`. In addition please let me know your opinion on this`try` `catch`, to potentially add it to other parts when calling `new URL()`.

nikostoulas: yeap orka has logger so we do it.

pmantafounis: added logger. Also added `try` `catch` in `isExpiredUrl()` bellow

nikostoulas (approved):

pmantafounis:
added export of joi interface so we can extend it or override it from the apps. CYPTAL again?

nikostoulas (approved):

PR #349, opened 2023-11-07
Joi: only evaluate s3HostBucketInPath if needed

pmantafounis:
small improvement: declare `s3HostBucketInPath` as a function to avoid unnecessary evaluation (extra work)

### DIFF
--- a/src/initializers/joi.ts
+++ b/src/initializers/joi.ts
@@ -14,11 +14,11 @@ export const isOwnS3Path = (bucket: string, val: string): boolean => {
     const { host, protocol, pathname } = new URL(val);
     const matchingProtocol = protocol === 'https:';
     const s3Host = host === `${bucket}.s3.amazonaws.com`;
-    const s3HostBucketInPath =
+    const s3HostBucketInPath = () =>
       host.startsWith('s3.') &&
       host.endsWith('.amazonaws.com') &&
       pathname.startsWith(`/${bucket}/`);
-    return matchingProtocol && (s3Host || s3HostBucketInPath);
+    return matchingProtocol && (s3Host || s3HostBucketInPath());
   } catch (e) {
     logger.error(`Failed to parse url: ${val}`, e);
     return false;


src/initializers/joi.ts
```diff
@@ -14,11 +14,11 @@ export const isOwnS3Path = (bucket: string, val: string): boolean => {
     const { host, protocol, pathname } = new URL(val);
     const matchingProtocol = protocol === 'https:';
     const s3Host = host === `${bucket}.s3.amazonaws.com`;
-    const s3HostBucketInPath =
+    const s3HostBucketInPath = () =>
       host.startsWith('s3.') &&
       host.endsWith('.amazonaws.com') &&
       pathname.startsWith(`/${bucket}/`);
-    return matchingProtocol && (s3Host || s3HostBucketInPath);
+    return matchingProtocol && (s3Host || s3HostBucketInPath());
```
nikostoulas, line 21: @pmantafounis to be sincere I am not sure if calling a function or always evaluating ` host.startsWith('s3.')` is better.
Can you run in a console the following to test?
```
profiler( ()=> isOwnS3Path(pathWithS3)

profiler(()=> isOwnS3Path(pathwiths3Host)
```
And compare the results for each solution with the above?

pmantafounis: Oops very nice point.
Indeed since only part of the statement is evaluated each time (when falsy) the previous implementation appears to is faster:
```
Function () => oldIsOwnS3Path('workable-application-form', 'https://workable-application-form.s3.amazonaws.com') ran 2,626,204 times in 998.22 ms
Function () => isOwnS3Path('workable-application-form', 'https://workable-application-form.s3.amazonaws.com') ran 2,483,601 times in 998.94 ms
```
My bad, thank you

nikostoulas: Was it also faster for the case of s3Host?

pmantafounis: Well I've actually profiled the case where `s3Host = true`, which is where the extra evaluation is done with `host.startsWith('s3.')`. When `s3Host = false`, `s3HostBucketInPath` will perform the evaluations in both solutions, so there was no need to profile (since the old one was faster even with the extras). However here are some example profile results:
```
Function () => isOwnS3Path('urlwheres3hostisfalse') ran 2,639,935 times in 998.20 ms
Function () => oldIsOwnS3Path( 'urlwheres3hostisfalsem') ran 2,750,769 times in 996.01 ms
```

nikostoulas: Nice thanks!

PR #353, opened 2023-12-01
Improved Kafka and RabbitMQ Functionality

nikostoulas:
## Description
This pull request includes several updates aimed at enhancing the functionality of Kafka and RabbitMQ. The changes include the deprecation of a Kafka function, propagation of headers in Kafka messages, updates to the examples configuration, and modifications to the logTracer and rabbitmq initializer functions.

## Changes
- Deprecated `kafka send message` in favor of `kafka.producer.send`
- Enabled propagation of headers in Kafka messages
- Updated examples configuration
- Modified `logTracer` function to return an empty string instead of undefined
- Updated `rabbitmq initializer` to return `originalTryHandle` call

### DIFF
--- a/src/initializers/kafka/base-kafka-handler.ts
+++ b/src/initializers/kafka/base-kafka-handler.ts
@@ -4,6 +4,8 @@ import { flatten } from 'lodash';
 import type * as KafkajsType from 'kafkajs';
 import { runWithContext } from '../../builder';
 import { logMetrics } from '../../helpers';
+import { appendToStore } from '../../utils';
+import type OrkaBuilderType from '../../orka-builder';
 
 export type BaseKafkaHandlerOptions = {
   topic: string | string[];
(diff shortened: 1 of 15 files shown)


src/initializers/kafka/base-kafka-handler.ts
```diff
@@ -72,24 +73,28 @@ abstract class Base<Input, Output> {
...
       topic,
       partition,
       ...(this.jsonParseValue && message.value
         ? { rawValue: message.value, value: this.parseValue(message.value) }
         : {}),
-      ...(this.stringifyHeaders && message.headers
-        ? { rawHeaders: message.headers, headers: this.parseHeaders(message.headers) }
-        : {})
+      rawHeaders: message.headers,
+      headers: this.parseHeaders(message.headers)
```
manoskouvarakis, line 97: As you removed condition for headers, do we need the same for rawValue above?

nikostoulas: Headers are buffers and toString() should always safe.
Kafka value could not be a json so parsing it to json would not be needed if the value someone uses is not json.
So no there is no need to remove it for value, as for binary for example content it would be an overkill to try json parse the messages.

manoskouvarakis (approved):

PR #358, opened 2024-01-04
[PROD-34066] Upgrades mongoose to version 7 and removes mongodb dependency

iliasbibas:
Added `mongoose@^7.6.7` (latest 7.x) and `mongodb@~5.9.1`.

The reason for upgrading `mongoose` version is [its official support for MongoDB 7.](https://mongoosejs.com/docs/compatibility.html)

Developers of apps using mongoose through orka should keep in mind [the migration guide](https://mongoosejs.com/docs/7.x/docs/migrating_to_7.html) for any breaking changes in the apps.


package.json
```diff
@@ -66,7 +66,8 @@
     "koa2-cors": "^2.0.6",
     "lodash": "^4.17.15",
     "log4js": "^6.7.1",
-    "mongoose": "6.*",
+    "mongodb": "~5.9.1",
```
jgerbe, line 69: ❓ why so strict? why not "^5.9.1"

package.json
```diff
@@ -66,7 +66,8 @@
...
+    "mongoose": "^7.6.7",
```
jgerbe, line 70: ❓ since orka is in 4.0 beta, do we want to make it only mongo7 compatible? @nikostoulas WDYT?

package.json
iliasbibas: I see in the [compatibility table of mongo](https://www.mongodb.com/docs/drivers/node/current/compatibility/#compatibility-table-legend) that they change support (upwards) using minor versions for the driver, so I opted to keep the minor version constant for orka as well. I changed it to `^` to be less restrictive.

iliasbibas:
Btw regarding the note in the PR description. I couldn't find a way to make `mongodb` be installed at the top level of the `node_module` (as it did with mongoose 6.x), although I expected it to be the default behaviour. If anyone has any suggestions on how to do it, please let me know so that we can remove the explicit `mongodb` dependency in the package.json.

nikostoulas:
@iliasbibas have you tried removing the dependency to mongodb altogether?

iliasbibas:
> @iliasbibas have you tried removing the dependency to mongodb altogether?

Working on it, I'll get back to you once done.

package.json
nikostoulas: Sure! Let's use 7.* to not restrict apps from getting latest updates

src/initializers/joi.ts
```diff
@@ -81,7 +81,7 @@ const Joi: JoiWithExtensions = _Joi.extend(
     base: joi.string().meta({baseType: 'string'}),
     messages: {'objectId.invalid': 'Invalid objectId'},
     validate: (value, helpers) => {
-      return require('mongodb').ObjectId.isValid(value)
+      return require('mongoose').isValidObjectId(value)
```
nikostoulas, line 84: 👏

nikostoulas (approved):

package.json
iliasbibas: 👍 I'll use the `7.*` notation for clarity

jgerbe (approved):

manoskouvarakis (approved):

PR #365, opened 2024-02-22
Add growthbook success log

pmantafounis:
Add a log upon successful loading of growthbook features. This informs the user that growthbook connection is established.

### DIFF
--- a/src/orka-builder.ts
+++ b/src/orka-builder.ts
@@ -166,7 +166,10 @@ export default class OrkaBuilder {
       const { createGrowthbook } = require('./initializers/growthbook');
       const gb = createGrowthbook(this.config.growthbook);
       if (!gb) return;
-      await gb.loadFeatures().catch(e => getLogger('orka.growthbook').error('Unable to load features', e));
+
+      const logger = getLogger('orka.growthbook');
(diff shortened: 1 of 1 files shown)


src/orka-builder.ts
```diff
@@ -166,7 +166,10 @@ export default class OrkaBuilder {
       const { createGrowthbook } = require('./initializers/growthbook');
       const gb = createGrowthbook(this.config.growthbook);
       if (!gb) return;
-      await gb.loadFeatures().catch(e => getLogger('orka.growthbook').error('Unable to load features', e));
+
+      const logger = getLogger('orka.growthbook');
+      await gb.loadFeatures().catch(e => logger.error('Unable to load features', e));
+      logger.info('Growthbook features loaded');
```
nikostoulas, line 172:
```suggestion
      await gb.loadFeatures()
      .then(()=>  logger.info('Growthbook features loaded'))
      .catch(e => logger.error('Unable to load features', e));
```

Sth like this ☝️ wouldn't log on errors

pmantafounis: Indeed, thanks. Changed it in [ccbd5f5](https://github.com/Workable/orka/pull/365/commits/ccbd5f588160c40dce5042cf8102f44f6b716062)

PR #367, opened 2024-03-04
Add postgres statement_timeout and query_timeout

manoskouvarakis:
We want to be able to configure statement and query timeout in a postgres pool. Without adding any variable, the default will remain to no timeout.

Currently in an application that uses orka's postgres connection, this can be handled only in database level. 

Documentation: https://node-postgres.com/apis/client

### DIFF
--- a/src/initializers/postgres.ts
+++ b/src/initializers/postgres.ts
@@ -20,7 +20,9 @@ export default function postgres(config) {
       max: pgConfig.poolSize,
       ssl: pgConfig.useSsl ? pgConfig.sslConfig : undefined,
       idleTimeoutMillis: pgConfig.idleTimeoutMillis,
-      connectionTimeoutMillis: pgConfig.connectionTimeoutMillis
+      connectionTimeoutMillis: pgConfig.connectionTimeoutMillis,
+      statement_timeout: pgConfig.statementTimeout,
+      query_timeout: pgConfig.queryTimeout,
     });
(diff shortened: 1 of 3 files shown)


src/initializers/postgres.ts
```diff
@@ -20,7 +20,9 @@ export default function postgres(config) {
       max: pgConfig.poolSize,
       ssl: pgConfig.useSsl ? pgConfig.sslConfig : undefined,
       idleTimeoutMillis: pgConfig.idleTimeoutMillis,
-      connectionTimeoutMillis: pgConfig.connectionTimeoutMillis
+      connectionTimeoutMillis: pgConfig.connectionTimeoutMillis,
+      statement_timeout: pgConfig.statementTimeout,
```
nikostoulas, line 24: Weird that they have different casing in the configuration

nikostoulas (approved):
Shall we update our docs too? See https://workable.github.io/orka/integrations/postgres.html

manoskouvarakis:
> Shall we update our docs too? See https://workable.github.io/orka/integrations/postgres.html

Added both options in documentation with the default value of no timeout: https://github.com/Workable/orka/pull/367/commits/89eaae133c6747c2e673a45e28bfef8afef7e94b

src/initializers/postgres.ts
manoskouvarakis: Indeed - I didn't like it also 🤯

PR #368, opened 2024-03-06
[PROD-37017] Allow orka to accept s3 presigned URLs with region

ChristosLabrou:
- Expands the `isOwnS3Path` function to accept presigned URLs that contain region in the hostname.


nikostoulas (approved):
👏

PR #379, opened 2024-04-29
[PROD-40813] Allow any tag on safeHtml

klesgidis:
## Symmary
* Allows passing `false` to rules

See https://www.npmjs.com/package/sanitize-html `"What if I want to allow all tags or all attributes?"`

### DIFF
--- a/src/initializers/joi.ts
+++ b/src/initializers/joi.ts
@@ -2,7 +2,7 @@ import * as _Joi from 'joi';
 import {isString} from 'lodash';
 import * as sanitizeHtml from 'sanitize-html';
 import {URL} from 'url';
-import { getLogger } from './log4js';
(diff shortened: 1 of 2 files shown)


src/initializers/joi.ts
```diff
@@ -2,7 +2,7 @@ import * as _Joi from 'joi';
 import {isString} from 'lodash';
 import * as sanitizeHtml from 'sanitize-html';
 import {URL} from 'url';
-import { getLogger } from './log4js';
+import {getLogger} from './log4js';
```
manoskouvarakis, line 5: ❓ Should those spaces be removed? Do you maybe have a different rule in linter because on node apps I remember keeping those spaces

klesgidis: I applied prettier to the whole file

PR #385, opened 2024-06-21
Pass host field name to riviere to be able to use custom host

manoskouvarakis:
After making host field name configurable in riviere `1.18.0` we should initialize it via orka in order to support custom host on all node applications.

https://workable.atlassian.net/browse/PROD-43442

### DIFF
--- a/src/initializers/diamorphosis.ts
+++ b/src/initializers/diamorphosis.ts
@@ -93,6 +93,7 @@ export default (config, orkaOptions: Partial<OrkaOptions>) => {
     styles: [],
     headersRegex: '(^X-.*)|cf-ray',
     maxBodyValueChars: undefined,
+    hostFieldName: 'host',
     ...config.riviere
   };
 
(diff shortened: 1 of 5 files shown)


src/initializers/diamorphosis.ts
```diff
@@ -93,6 +93,7 @@ export default (config, orkaOptions: Partial<OrkaOptions>) => {
     styles: [],
     headersRegex: '(^X-.*)|cf-ray',
     maxBodyValueChars: undefined,
+    hostFieldName: 'host',
```
manoskouvarakis, line 96: @nikostoulas I will use current host as the default value. Wdyt? 
In order to support DD all applications will have to set it in config for now that we are still in stackdriver production

nikostoulas: Makes sense. We should make a commit with SREs for all apps in k8s to add the respective env variable.

nikostoulas (approved):

src/initializers/diamorphosis.ts
manoskouvarakis: As we will have to bump orka (and riviere) in all apps, we can also add it there in codebase

PR #398, opened 2025-02-04
Set kafka authentication timeout to 10s (default)

manoskouvarakis:
As per documentation, kafkajs authenticationTimeout default value is 10s: https://kafka.js.org/docs/configuration#sasl

### DIFF
--- a/test/diamorphosis/diamorphosis.test.ts
+++ b/test/diamorphosis/diamorphosis.test.ts
@@ -92,7 +92,7 @@ describe('Diamorphosis Test', () => {
             sasl: { mechanism: '', username: '', password: '' }
           },
           connectionTimeout: 1000,
-          authenticationTimeout: 1000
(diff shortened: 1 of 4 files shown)


test/diamorphosis/diamorphosis.test.ts
```diff
@@ -92,7 +92,7 @@ describe('Diamorphosis Test', () => {
             sasl: { mechanism: '', username: '', password: '' }
           },
           connectionTimeout: 1000,
```
nikostoulas, line 94: ❓ Shall we bump connectionTimeout too to 3000?

nikostoulas (approved):

test/diamorphosis/diamorphosis.test.ts
manoskouvarakis: The default is 1000. I think we can do it on application level. Wdyt?

PR #399, opened 2025-02-10
PROD-54189 Convert Riviere's config blackistedPathRegex to RegExp when passing it

ChristosLabrou:
Converted from https://github.com/Workable/riviere/pull/113

We discovered that when `blacklistedPathRegex` was set via env var it would be passed as string to Riviere instead of a RegExp object which caused requests to not get logged.

### DIFF
--- a/src/initializers/riviere.ts
+++ b/src/initializers/riviere.ts
@@ -24,11 +24,11 @@ const init = (config, orkaOptions) => {
       https: true,
       level: 'info',
       maxBodyValueChars: config.riviere.maxBodyValueChars,
-      blacklistedPathRegex: config.riviere.outbound && config.riviere.outbound.blacklistedPathRegex,
       request: {
         enabled: config.riviere.outbound && config.riviere.outbound.request.enabled
       },
-      ...config.riviere.outbound
+      ...config.riviere.outbound,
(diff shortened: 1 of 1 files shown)


src/initializers/riviere.ts
```diff
@@ -24,11 +24,11 @@ const init = (config, orkaOptions) => {
       https: true,
       level: 'info',
       maxBodyValueChars: config.riviere.maxBodyValueChars,
-      blacklistedPathRegex: config.riviere.outbound && config.riviere.outbound.blacklistedPathRegex,
       request: {
         enabled: config.riviere.outbound && config.riviere.outbound.request.enabled
       },
-      ...config.riviere.outbound
+      ...config.riviere.outbound,
+      blacklistedPathRegex: config.riviere.outbound && new RegExp(config.riviere.outbound.blacklistedPathRegex, 'i')
```
klesgidis, line 31: like we said, if someone does not override the value from helm release and set it directly to `config` with a regex it will fail

ChristosLabrou: It doesn't fail when testing it. Passing a RegExp object to the RegExp returns a RegExp object with the same `source` key

Example:
```ts
const r1 = new RegExp('some-regex.*');
console.log(r1.source); // prints 'some-regex.*'
const r2 = new RegExp(r1)
console.log(r2.source) // also prints 'some-regex.*'
```

nikostoulas (approved):

PR #402, opened 2025-02-13
Update regex handling to respect flags from config

ChristosLabrou:
Currently when Orka is passing a regex (to riviere), it overwrites any flags that were set in the app config. This PR changes that so it will respect flags that were set in config when overriding a value from config/env values.

### DIFF
--- a/src/initializers/riviere.ts
+++ b/src/initializers/riviere.ts
@@ -28,7 +28,7 @@ const init = (config, orkaOptions) => {
         enabled: config.riviere.outbound && config.riviere.outbound.request.enabled
       },
       ...config.riviere.outbound,
-      blacklistedPathRegex: config.riviere.outbound && new RegExp(config.riviere.outbound.blacklistedPathRegex, 'i')
+      blacklistedPathRegex: config.riviere.outbound && RegExp(config.riviere.outbound.blacklistedPathRegex)
     },
     inbound: {
       level: 'info',
@@ -42,11 +42,11 @@ const init = (config, orkaOptions) => {
     errors: {
       enabled: config.riviere.enabled
     } as any,
-    headersRegex: new RegExp(config.riviere.headersRegex, 'i'),
+    headersRegex: RegExp(config.riviere.headersRegex),
     traceHeaderName: config.traceHeaderName,
     styles: config.riviere.styles,
     bodyKeys: config.riviere.bodyKeys,
-    bodyKeysRegex: config.riviere.bodyKeysRegex && new RegExp(config.riviere.bodyKeysRegex, 'i'),
+    bodyKeysRegex: config.riviere.bodyKeysRegex && RegExp(config.riviere.bodyKeysRegex),
     bodyKeysCallback: config.riviere.bodyKeysCallback,
     color: config.riviere.color,
     hostFieldName: config.riviere.hostFieldName,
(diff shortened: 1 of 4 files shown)


src/initializers/riviere.ts
```diff
@@ -28,7 +28,7 @@ const init = (config, orkaOptions) => {
         enabled: config.riviere.outbound && config.riviere.outbound.request.enabled
       },
       ...config.riviere.outbound,
-      blacklistedPathRegex: config.riviere.outbound && new RegExp(config.riviere.outbound.blacklistedPathRegex, 'i')
+      blacklistedPathRegex: config.riviere.outbound && RegExp(config.riviere.outbound.blacklistedPathRegex)
```
ChristosLabrou, line 31: JS docs mention this RegExp constructor feature:

> RegExp(pattern) returns pattern directly if all of the following are true:
> RegExp() is called without new
> pattern.constructor === RegExp (usually meaning it's not a subclass);
> flags is undefined.

We use that feature when passing config values to avoid overriding flags with our hardcoded one.

[Link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/RegExp#return_value) for full docs

nikostoulas: ❓ Since diamorphosis now handles regexes could we simplify this as: 
```ts
blacklistedPathRegex: config.riviere.outbound. blacklistedPathRegex
```
?
Since this is done in ...config.riviere.outbound we could possibly remove this line altogether

ChristosLabrou: Done, it was possible to simplify after all.

src/initializers/riviere.ts
```diff
@@ -42,11 +41,11 @@ const init = (config, orkaOptions) => {
     errors: {
       enabled: config.riviere.enabled
     } as any,
-    headersRegex: new RegExp(config.riviere.headersRegex, 'i'),
+    headersRegex: RegExp(config.riviere.headersRegex),
```
nikostoulas, line 44: ❓ Could this also be done:
```
headersRegex: config.riviere.headersRegex
```
?

nikostoulas (approved):
👏

src/initializers/riviere.ts
ChristosLabrou: Done. Also fixed `bodyKeysRegex` to do the same

src/initializers/riviere.ts
```diff
@@ -11,6 +11,25 @@ const https = require('https');
 let middleware;
 
 const init = (config, orkaOptions) => {
+  // TODO: Remove this conversion logic after v5.x.x release
```
ChristosLabrou, line 14: We had a discussion with @nikostoulas about this.
Some of our apps use strings in their config files. We've decided to add a warning and conversion to be backwards compatible. It should be removed in the next major version release.

src/initializers/riviere.ts
```diff
@@ -11,6 +11,25 @@ const https = require('https');
...
+  const logger = getLogger('orka.riviere');
+  const regexKeys = ['headersRegex', 'bodyKeysRegex'];
+  regexKeys.forEach(key => {
+    if (typeof config?.riviere[key] === 'string') {
+      logger.warn(
+        `You are using a string for regex key ${key} in riviere config. This will not supported after Orka v5.x.x. Please use a RegExp object.`
```
nikostoulas, line 20:
```suggestion
        `You are using a string for regex key ${key} in riviere config. This will not be supported after Orka v5.x.x. Please use a RegExp object.`
```

nikostoulas (approved):

PR #405, opened 2025-04-15
PROD-57294: Reuse Redis connections for Bull

kkyriakidis:
Bull is currently using 3 connections per Bull queue
https://github.com/OptimalBits/bull/blob/develop/PATTERNS.md

With this implementation we give the option to reuse the 'client' and 'subscriber' connection across different Bull queues


src/initializers/bull/bull.ts
```diff
@@ -3,6 +3,7 @@ import { getLogger } from '../log4js';
 import requireInjected from '../../require-injected';
 import Prometheus from '../prometheus/prometheus';
 import { JobCounts } from 'bull';
+import IORedis = require('ioredis');
```
Nikoklis, line 6: ❓ Don't we use `redis` package for connecting to redis

nikostoulas (approved):

src/initializers/bull/bull.ts
```diff
@@ -15,25 +16,59 @@ export default class Bull {
...
+            });
+          default:
+            throw new Error('Unexpected connection type: ' + type);
+        }
+      }
+    };
     const queueOptions = { redis: this.redisOpts, defaultJobOptions, ...(limiter && { limiter }) };
     this.logger.info(`Creating Queue: ${fullName}`);
-    const queue = new Queue(fullName, queueOptions);
+    const queue = new Queue(fullName, this.reuseClients ? _.defaultsDeep(queueOptions, reuseOptions) : queueOptions);
```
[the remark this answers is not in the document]
kkyriakidis: sory Niko. Added with the second push

src/initializers/bull/bull.ts
Nikoklis: I got confused with Orka dependency and bull dependency here. sorry you are right 👍

src/initializers/bull/index.ts
```diff
@@ -17,6 +17,7 @@ export default async (config, appName: string) => {
   const redis = config.bull.redis || config.redis;
   const options = { enableReadyCheck: false, ...omit(config.bull.redis || config.redis.options, ['url', 'tls']) };
   const queues = config.bull.queue.queues;
+  const reuseConnections = config.bull.reuseConnections;
```
nikostoulas, line 20: ❓ Shall we add a default value in diamorphosis.ts?

PR #409, opened 2025-05-14
Switch from bull to bullmq

jgerbe:
this is a breaking change


nikostoulas (approved):
Nide @jgerbe . Since this is a breaking change let's add in the Readme a v4.xx to v5.xx section with breaking changes. The next version we will built should be a preview version of `v5.0.0-beta.1` until we put all the breaking changes we want into v5. 

@konkyrkos has another breaking change in mind too.

iliasbibas (approved):

PR #412, opened 2025-06-23
Add original Axios response to AxiosError of orka

nikoskoukoulas:
The issue:
Axios Interceptor changes the original type of the AxiosError. As a result packages that depend on axios and happen to be on the same version get a modified version of Error. Most of this packages like [helloSign](https://github.com/hellosign/dropbox-sign-node/blob/main/api/signatureRequestApi.ts#L245) depend on the response to handle the error. With this pr this is feasible.


stavros-wb (approved):
lgtm

vboutas (approved):

PR #503, opened 2026-08-06
Migrate redis from v3 to v6

Nikoklis:
Replaces #490. That bump cannot be merged as-is: node-redis v4 rewrote the client API and the build fails. This PR does the actual migration.

## Before / after

**Client usage (breaking for orka consumers):**

```js
// before (v3, callbacks)
getRedis().get('key', (err, value) => { ... });

// after (v6, promises)
const value = await getRedis().get('key');
```

**`config.redis` schema: unchanged.** `url`, `options.tls` and the legacy retry/keepalive options (`timesConnected`, `totalRetryTime`, `reconnectAfterMultiplier`, `socketKeepalive`, `socketInitialDelay`) keep working and are mapped to the new driver options. Any other key in `config.redis.options` (e.g. `RESP`, `pingInterval`, `socket`) is passed through to `createClient`.

**Initializer internals:**

| | before (v3) | after (v6) |
|---|---|---|
| Client creation | `createClient(url, options)` | `createClient({ url, RESP: 2, socket })` |
| Connection | implicit | explicit `client.connect()`, awaited during boot; if redis is down the app still boots and `/health` reports unhealthy |
| Reconnect | `retry_strategy` | `socket.reconnectStrategy`, same semantics |
| Too many reconnects | thrown error crashed the process | stops reconnecting; `/health` turns 503 so the orchestrator restarts the pod |
| Health check | `client.connected` | `client.isReady` |
| Keep-alive | manual `setKeepAlive` on `client.stream` | `socket.keepAlive` / `socket.keepAliveInitialDelay` |
| Close (tests) | `client.end(true)` | `client.destroy()` |
| Types | `@types/redis` | shipped with the driver (`@types/redis` removed) |

`RESP` is pinned to `2` to preserve reply shapes; set `config.redis.options.RESP = 3` to opt in to RESP3.

## Breaking changes

- `getRedis()` returns a promise-based node-redis v6 client; callback-style commands and `client.end()` are gone.
- `createRedisConnection()` is now async and resolves after the initial connection attempt.
- node-redis v6 requires Node >= 20 (CI matrix already is 20/22/24).
- Requires a major orka release.

## Testing

- Unit tests rewritten for the new API (option mapping, TLS, all reconnect-strategy branches).
- Redis example e2e passes against a real redis: connect, `/health` 200, `destroy()`, `/health` 503.
- Manual smoke test of `set`/`get`/`del` through `getRedis()`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ch-asimakopoulos (approved):

iliasbibas (approved):

klesgidis (approved):
LGTM! maybe we should give it a try before releasing this

